PR #182, opened 2018-03-28
Feature/locate objects

Jeffroiscool:
First iteration, will need a center function for 3D just like in isometric but the camera system is a bit challenging for me.

I'd like to expand on this by making hidden models/bounding boxes and the name tag visible of whatever I choose to locate.

### DIFF
--- a/src/game/loop/cameras/iso.js
+++ b/src/game/loop/cameras/iso.js
@@ -27,7 +27,7 @@ export function processFreeIsoMovement(controlsState, camera, time) {
 }
 
 function getObjectIsoPos(renderer, camera, object) {
-    const objectHeight = 0;
+    var objectHeight = 0;
     if(object.model){
         const bb = object.model.boundingBox;
         objectHeight = bb.max.y - bb.min.y;
--- a/src/game/loop/cameras/iso.js
+++ b/src/game/loop/cameras/iso.js
@@ -26,7 +26,7 @@ export function processFreeIsoMovement(controlsState, camera, time) {
 }
 
 function getObjectIsoPos(renderer, camera, object) {
-    var objectHeight = 0;
+    let objectHeight = 0;
     if (object.model){
         const bb = object.model.boundingBox;
         objectHeight = bb.max.y - bb.min.y;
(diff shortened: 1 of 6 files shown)


src/game/loop/cameras/iso.js
```diff
@@ -24,12 +26,16 @@ export function processFreeIsoMovement(controlsState, camera, time) {
     camera.updateProjectionMatrix();
 }
 
-function getActorIsoPos(renderer, camera, actor) {
-    const bb = actor.model.boundingBox;
-    const actorHeight = bb.max.y - bb.min.y;
-    const pos = new THREE.Vector3(0, actorHeight * 0.5, 0);
-    actor.threeObject.updateMatrixWorld();
-    pos.applyMatrix4(actor.threeObject.matrixWorld);
+function getObjectIsoPos(renderer, camera, object) {
```
agrande: I think we should return early on this if there's no model in the object, and not use a default height.

src/game/loop/cameras/iso.js
```diff
@@ -24,12 +26,16 @@ export function processFreeIsoMovement(controlsState, camera, time) {
...
+    const objectHeight = 0.04;
+    if(object.model){
```
agrande: I think we should return early on this if there's no model in the object, and not use a default height.

src/ui/editor/DebugData.js
```diff
@@ -126,6 +127,16 @@ export function getObjectName(type, sceneIndex, objIndex) {
     return `${type}${objIndex}`;
 }
 
+export function locateObject(varDef){
+    console.log(varDef);
```
agrande: Remove the console.log() after you tested the code.

Jeffroiscool: Oops, forgot about that. Is it automatically removed now?

src/game/loop/cameras/iso.js
agrande: Ok, so since this is used for objects that don't have a model attached too, I think instead the default height should be 0.

src/ui/editor/DebugData.js
agrande: What do you mean automatically?

agrande:
Also, when selecting Zoe2 in Twinsen's house, I get a crash: `Uncaught TypeError: Cannot read property 'updateMatrixWorld' of null`

PR #270, opened 2020-03-09
Feature/support open hqr format for voices

pleasenophp:
Introduced OpenHqr format, which is zip packed HQR with header in JSON format and extracted files. The format also supports both indexes and names (for future).

The hqr.ts now supports both Hqr and OpenHqr formats. This is for now is enabled for voices. So voice loading code is trying to load first .VOX and if not found, then loads .VOX.zip

hqr.ts is easily configurable to try to load OpenHqr first, then HQR, other way around, or only one of them. By default hqr.ts still tries to load only HQR. Trying to load OpenHqr if HQR not found is enabled only for voices for now.

npm run convert voice [bitrate] - will now convert all the default game VOX files to OpenHqr format with packed AAC files. This command ignores *_AAC.vox files (thus are manually converted vox, they are not part of original game content). 

Example:
`npm run convert voice 64`

Will convert all voices to OpenHqr with 64k bitrate. 

Other than that, any HQR can now be repacked to OpenHqr using 'npm run convert hqr' command. This command is not to use for end user, just for internal usage. This command doesn't do any converting on items, just saves them as they are. It's not designed for VOX files, for VOX use npm run convert voice instead.

Usage
`npm run convert hqr <path to HQR file>`

Example:
`npm run convert hqr ./www/data/BODY.HQR`

Will result in ./www/data/BODY.HQR.zip

Also, fixed bug in hqr reader code, where not all hidden entries were always read. Updated the wiki with more precise hidden entries specification. 
http://lbafileinfo.kazekr.net/index.php?title=High_quality_resource


xesf:
Everything looks fine in the PR, I just have a doubt in having the preload of the original VOX files there as well. We probably just need to decide and use the OpenHQR everywhere across the application.

It will also help some of the hard-coded reference to HQR files prefect in code, but that's for another PR. I was thinking something that could not point directly to a file name but instead to a resource type which can vary between LBA versions.

pleasenophp:
Yes, you can enable OpenHqr only later for all the files if you want. For now I want to make the audio and samples be converted automatically, and work for users. I want the existing users to still be able to work with old _AAC.vox files without forcing them to run convertors again. When you think it's time, you can completely switch to the new format.

src/game/index.ts
```diff
@@ -133,6 +133,10 @@ export function createGame(clock: any, setUiState: Function, getUiState: Functio
                 preloadFile('data/RESS.HQR', tr('Resources')),
                 preloadFile(`data/VOX/${languageVoice.code}_GAM_AAC.VOX`, tr('MainVoices'), false),
                 preloadFile(`data/VOX/${languageVoice.code}_000_AAC.VOX`, tr('Voices'), false),
+                preloadFile(`data/VOX/${languageVoice.code}_GAM_AAC.VOX.zip`,
+                    tr('MainVoices.zip'), false),
+                preloadFile(`data/VOX/${languageVoice.code}_000_AAC.VOX.zip`,
+                    tr('Voices.zip'), false),
```
agrande: tr() is used for translating strings, I don't think it's needed here for a file name...

src/audio/index.ts
```diff
@@ -285,7 +285,7 @@ function getVoiceSource(state, context, data = null) {
         if (textBankId === -1) {
             filename = `VOX/${state.config.languageVoice.code}_GAM_AAC.VOX`;
         }
-        loadHqr(filename, true).then((voices) => {
+        loadHqr(filename, true, [HqrFormat.HQR, HqrFormat.OpenHQR]).then(async (voices) => {
```
agrande: I'm not sure to understand why we need to try both formats.
Why can't we just pick one or the other, and autodetect with the file extension by default?

agrande:
> Yes, you can enable OpenHqr-only later for all the files if you want. For now I want to make the audio and samples be converted automatically, and work for new users. I want the existing users and the demo website to still be able to work with old _AAC.vox files without forcing them to run convertors. When you think it's time, you can completely switch to the new format.

It's not like we have tons of users, we're basically just 4 people actively (more or less) working on the project at this point... So we can just go for instantly switching to the new version, and I'll switch the demo website data folder on the next release.

src/webapi.ts
```diff
@@ -0,0 +1,35 @@
...
+            request.onerror = function onerror(err) {
+                resolve({
+                    status: -1,
+                    error: err,
+                });
+            };
+            request.send(null);
+        });
+    }
+}
```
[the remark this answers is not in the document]
pleasenophp: I just needed to extract HTTPRequest code from hqr.ts to avoid duplication. Other code also uses directly the HTTPRequest so could be changed to use same API. I don't want to experiment with using new API in this PR. But you can change it to any other API ofc. The idea is to create some abstraction point for all the web calls, so it can be changed in one place.

src/audio/index.ts
pleasenophp: I don't want to break backwards compatibility with this, but if you say it's easy for you to switch to new format, I will just change it to only try OpenHQR.

src/webapi.ts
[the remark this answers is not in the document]
agrande: The thing is, fetch does pretty much the same abstraction, and it is a web standard that also has node.js implementations, so that would be a nice replacement. I didn't know about it at the time I started the project so I went for raw HTTPRequests and stuck with that, and then started to use async/await and never went for a full refacto.
I suppose I'll merge the PR like this and replace everything with fetch later on.

src/audio/index.ts
agrande: Well, I think backwards compatibility adds code complexity that we don't really need.

src/webapi.ts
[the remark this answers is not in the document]
pleasenophp: Yeah, I suppose it's good idea to do it separately as switching to fetch is not part of this story.

PR #284, opened 2020-06-01
Stop theme music playing when exiting VR.

slwilliams:
Without this we have two copies playing because we never clean up the VR state.


slwilliams:
Lint failure is unrelated (existing line length issue I'm not touching).

agrande:
Well, the lint error happens on line 301 => 306 which you did change.
Doing the bind(this) in the constructor would remove the issue, in addition to avoiding rebinding at each render() call.

agrande:
Besides that, this PR looks good to me. Thanks!

slwilliams:
Ah ok yea, done.

PR #285, opened 2020-06-02
Make the b button on VR controllers open the menu.

slwilliams:


slwilliams:
Actually I still need to add the help text when the initial controls are rendered.

slwilliams:
Ok gone with "Menu", let me know if "Back" would be a better term, or something else.

src/controls/vr/mappings.ts
```diff
@@ -193,12 +199,12 @@ function mapMainGameControls(
         if (numControllers === 1 || handedness === 'right') {
             nextBehaviour = {
                 btn: 'xr-standard-squeeze',
-                handler: handleTapped
+                handler: handleTapped.bind({})
```
agrande, line 202: Damn! Seems like I forgot the binding...
I don't know how it could work without it.

agrande:
> Ok gone with "Menu", let me know if "Back" would be a better term, or something else.

I guess it depends on the perspective. The button does actually go back depending on where you are. But if you consider that the labels are showing the controls for when you're playing the game, then "Menu" makes more sense indeed.

Besides, I was initially planning to display the labels before starting the game when you start a new game, and when you press a specific "help" button in the UI, instead of always showing them in the menu. I didn't get around to it yet. And I'm still not entirely sure what's the best thing to do.

PR #287, opened 2020-06-05
Improve the Twinsen placement logic when transitioning from one scene to another.

slwilliams:
There is some reasonable complexity here so I've done my best to structure it in a nice way and comment it where appropriate.

Basic idea is:
  * Find where the zone.props coordinates would put us by default.
  * Find the closest zone in the new scene to those coordinates (this is the door we're exiting from).
  * If a closest zone doesn't exist, fallback to the default zone.props location.
  * If a closest zone does exist, snap our target location in the new scene to the closest zone bounding box.
  * Apply an offset defaulting to half a brick but allow overriding per scene transition for problematic cases.
  * Apply the delta of Twinsen compared to the door he's entering (i.e. how far along the door he is).

This implementation performs much better that what we currently have. I've spent some time testing and have identified a few offsets for cases which don't work. There are certainly more of these todo but I don't want to spend the time hunting them all down now.

### DIFF
--- a/src/game/loop/zones.ts
+++ b/src/game/loop/zones.ts
@@ -30,7 +30,7 @@ const SCENE_ID_TO_NAME = {
     109: 'FRANCO_VILLAGE',
     174: 'FRANCO_NURSERY',
     175: 'FRANCO_ROGER_HOUSE',
-}
+};
 
 // Map from (previous zone, target zone) to offset in bricks.
 // E.g. ZONE_OFFSET_OVERRIDES[CELLAR][TWINSENS_HOUSE] is the offset we should
@@ -38,17 +38,17 @@ const SCENE_ID_TO_NAME = {
 // from the CELLAR.
 // TODO: There are likely more of these needed.
 const ZONE_OFFSET_OVERRIDES = {
-    'TWINSENS_HOUSE': {
-        'CELLAR': 1.25,
+    TWINSENS_HOUSE: {
+        CELLAR: 1.25,
     },
-    'DESERT_ISLAND_TOWN_SQUARE': {
-        'DESERT_ISLAND_TICKET_SHOP': -1.25,
+    DESERT_ISLAND_TOWN_SQUARE: {
+        DESERT_ISLAND_TICKET_SHOP: -1.25,
     },
-    'FRANCO_VILLAGE': {
-        'FRANCO_NURSERY': -1.25,
-        'FRANCO_ROGER_HOUSE': -1.25,
+    FRANCO_VILLAGE: {
+        FRANCO_NURSERY: -1.25,
+        FRANCO_ROGER_HOUSE: -1.25,
     }
-}
+};
 
 export function processZones(game, scene) {
     const hero = scene.actors[0];
@@ -90,8 +90,8 @@ function debugZoneTargetPos(newScene, pos, color) {
 
 function getDistance(pos1, pos2) {
     return Math.sqrt(
-        Math.pow(pos1.x - pos2.x, 2) + 
-        Math.pow(pos1.y - pos2.y, 2) + 
+        Math.pow(pos1.x - pos2.x, 2) +
+        Math.pow(pos1.y - pos2.y, 2) +
         Math.pow(pos1.z - pos2.z, 2)
     );
 }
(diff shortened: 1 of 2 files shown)


src/game/loop/zones.ts
```diff
@@ -45,36 +75,146 @@ export function processZones(game, scene) {
...
 */
 
+function getDistance(pos1, pos2) {
+    return Math.sqrt(
+        Math.pow(pos1.x - pos2.x, 2) +
+        Math.pow(pos1.y - pos2.y, 2) +
+        Math.pow(pos1.z - pos2.z, 2)
+    );
+}
+
```
agrande, line 98: Why not use some three.js Vector3s here? No need to reimplement a custom getDistance function...

src/game/loop/zones.ts
```diff
@@ -45,36 +75,146 @@ export function processZones(game, scene) {
...
+            // Work out which orientation the zone is, depending on this we can
+            // determine which axis should be used to work out how far "along
+            // the zone" Twinsen is.
+            const lenX = zone.props.box.xMax - zone.props.box.xMin;
+            const lenZ = zone.props.box.zMax - zone.props.box.zMin;
+            if (lenX > lenZ) {
+                delta = (hero.physics.position.x - zone.props.box.xMin) / lenX;
+            } else {
+                delta = (hero.physics.position.z - zone.props.box.zMin) / lenZ;
+            }
```
agrande, line 164: Looks good for the orientation, but how do you work out the direction of the zone (which side of the zone is the actual door)?

src/game/loop/zones.ts
```diff
@@ -45,36 +75,146 @@ export function processZones(game, scene) {
...
+    );
+}
+
 /**
  * @return {boolean}
  */
 function GOTO_SCENE(game, scene, zone, hero) {
     if (!(scene.sideScenes && zone.props.snap in scene.sideScenes)) {
-        const box = zone.props.box;
         scene.goto(zone.props.snap).then((newScene) => {
```
agrande, line 104: The goto function has gotten pretty big here. Maybe we should create/externalize the function that handles the target zone pos guessing? Otherwise the rest of the transition logic gets burried into that code.

src/game/loop/zones.ts
```diff
@@ -45,36 +75,146 @@ export function processZones(game, scene) {
...
+                }
 
-            // debugZoneTargetPos(newScene, newHero);
+                // Again work out which orientation the zone is in the new scene
+                // allowing us to know which axis to apply the delta to.
+                const newBox = closestZone.props.box;
+                const newLenX = newBox.xMax - newBox.xMin;
+                const newLenZ = newBox.zMax - newBox.zMin;
+                if (newLenX > newLenZ) {
+                    if (initialTargetPos.z <= newBox.zMin) {
```
agrande, line 191: Ok, I guess that's where you determine the direction.

src/utils/lba.ts
```diff
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
 
+export const BRICK_SIZE = 512;
```
agrande, line 3: In what unit is that? Surely not meters.

src/game/loop/zones.ts
slwilliams: Yea good point.

src/game/loop/zones.ts
slwilliams: Added a comment to make this clearer.

src/game/loop/zones.ts
slwilliams: Agree, factored out into a separate function.

src/game/loop/zones.ts
slwilliams: Yea added a comment about this logic.

src/utils/lba.ts
slwilliams: This is "game units" it's what the zone.props.info0 etc. are in. 512 of those is one in game "brick".

slwilliams: Added a comment to explain.

slwilliams:
Addressed comments, PTAL.

PR #288, opened 2020-06-07
Add configurable offset for transparent objects per island.

slwilliams:
Transparent objects are placed quite far away from the models they're supposed to be stuck to. E.g. the symbols on the doors on Citadel island. Add a configurable offset per island to push these back to be closer to the object they're supposed to be attached too.

This makes things look much better.

This is actually a problem in the original game, it's just the camera system there doesn't let you get as close as ours and so never looked as bad. However given we allow playing in VR it's much more of a problem for us.

### DIFF
--- a/src/island/objects.ts
+++ b/src/island/objects.ts
@@ -5,13 +5,21 @@ import {WORLD_SCALE, WORLD_SIZE} from '../utils/lba';
 
 const push = Array.prototype.push;
 
-export function loadObjects(section, geometries, models, atlas) {
+// Offset amount per island to offset any transparent objects to be closer to
+// the object they're supposed to be attached to. This offset is unfortunately
+// not consistent between islands, although appears to be consistent within a 
+// given island.
+const TransparentObjectOffset = {
+    'CITADEL': 0.05,
+}
+
+export function loadObjects(section, geometries, models, atlas, island) {
     const numObjects = section.objInfo.numObjects;
     const boundingBoxes = [];
     for (let i = 0; i < numObjects; i += 1) {
         const info = loadObjectInfo(section.objects, section, i);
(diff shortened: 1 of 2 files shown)


src/island/objects.ts
```diff
@@ -5,13 +5,21 @@ import {WORLD_SCALE, WORLD_SIZE} from '../utils/lba';
 
 const push = Array.prototype.push;
 
-export function loadObjects(section, geometries, models, atlas) {
+// Offset amount per island to offset any transparent objects to be closer to
+// the object they're supposed to be attached to. This offset is unfortunately
+// not consistent between islands, although appears to be consistent within a
+// given island.
```
agrande, line 11: @slwilliams Does it work with the palm trees for instance? There are quite a few transparent objects in the island.

src/island/objects.ts
```diff
@@ -88,10 +96,21 @@ function loadSection(geometries, object, info, section, boundingBoxes, atlas) {
                 );
                 geometries.objects_colored.colors.push(getColor(section, i));
             } else {
+                const pos = getPosition(object, info, index);
+                if (section.isTransparent && TransparentObjectOffset[island.name]) {
+                    const normalVec = new THREE.Vector3(
```
agrande, line 101: @slwilliams carefull with allocating Vectors in a loop that's called often. It's a performance killer.
It's uglier to read, but in those cases, I would usually put a vector variable outside the function (with a name in UPPERCASE), and set it to a different value at each iteration (ex in the cameras update code for instance, or physics updates).

agrande: Anyway, I have to make an optimization pass on the loading code, I'm sure we can easily divide loading times by 10.

PR #294, opened 2020-06-12
Fixed Menu Music playing during Intro video due to a late load.

xesf:


src/ui/GameUI.tsx
```diff
@@ -256,7 +256,11 @@ export default class GameUI extends FrameListener<GameUIProps, GameUIState> {
     showMenu(inGameMenu = false) {
         this.state.game.pause();
         const audioMenuManager = this.state.game.getAudioMenuManager();
-        audioMenuManager.getMusicSource().loadAndPlay(6);
+        audioMenuManager.getMusicSource().load(6, () => {
+            if (this.state.showMenu) {
```
slwilliams, line 260: Perhaps add a comment explaining why this is needed over the simpler loadAndPlay?

E.g. "Ensure we don't start the music if it took a while to load the audio file and the user has already made it past the menu".

xesf: Cheers.
We tend to reduce the amount of comments in the code, so we can have everything more explicit as possible. Although, I will be refactoring the whole Audio part and improve this piece soon.

xesf: Pushed change with the comment.

xesf:
Closing PR as I have another one in-progress.

agrande:
@xesf can I delete the branch then, or do you prefer to keep it opened for some time?

xesf:
@agrande That's done. I have removed the branch. Thanks for the heads-up.

PR #298, opened 2020-06-18
Implement climbing up ladders

slwilliams:
I've had to introduce a few new things here:
  * The ability to disable keyframe interpolation for animations. The transition from climbing up to top out is exact and the interpolation was messing things up. Add a state option to disable this.
  * Add a new runtime flag to signify if Twinsen is colliding with a wall. Gate starting the ladder climbing based on Twinsen actually running into it.
  * Add the ability to execute a callback after an animation has played a single time. This allows us to exit climbing once the top out animation has played because by that point you're no longer in the ladder zone and so can't control anything from that code anymore. There might be a better way of doing this though.

I couldn't work out how to identify if Twinsen was facing ladders on islands because the zone box can be at an arbitrary angle, left a TODO for this.

This also doesn't implement climbing down backwards yet (but who ever did that anyways :D)


agrande:
I didn't even remember you could climb down ladders...

PR #299, opened 2020-06-18
Fix ladder anim

agrande:
@slwilliams The animation you used was actually for climbing down, and the speed you used was compensating the movement from the animation (but was framerate dependent)


src/game/loop/zones.ts
```diff
@@ -109,8 +108,6 @@ function LADDER(game, scene, zone, hero) {
                 return false;
```
slwilliams, line 108: We no longer need this block at all now because the upwards effect is in the animation!

src/game/data/animType.ts
```diff
@@ -13,7 +13,7 @@ export const AnimType = {
     PUNCH_1: 17,
     RUNNING_JUMP: 25,
     TALK: 28,
-    CLIMB_UP: 30,
+    CLIMB_UP: 12,
```
slwilliams, line 16: Re sort this list in index order.

PR #303, opened 2020-06-21
Fixed missing extra runtimeFlags as processCollision is shared betwee…

xesf:
…n different objects


xesf:
If you use the extra bonus and when the collision with the ground happened

PR #304, opened 2020-06-22
Fix implementation of getDistFromFloor

slwilliams:
Many bugs here:
  * getPositions should have been returning an array of Vector3!
  * Create an originalPos variable to use when calculating the distance to return - obj.physics.position doesn't have the matrixWorld applied so we shouldn't use that.
  * The entire structure of the while loop was messed up, restructure things to it works properly.

Some how even with all those (quite major tbh) issues, it still managed to work well enough for me to test and not find any issues?!


slwilliams:
Ack yea I did think of this when writing that line, but decided if we're likely to swap the physics engine out somewhat soon probably not worth worrying about it. Don't think we'll need this implementation for anything else except Twinsen for the foreseeable future.

PR #306, opened 2020-06-27
Fix ladders in VR.

slwilliams:
Turned out to be trivial, the control vector is just not exactly 1 for the thumbstick in VR. Be a bit more lenient with how we check for this.


agrande:
Oh I see, nice!

PR #310, opened 2020-07-04
Initial implementation for drowning.

slwilliams:
This is only an initial implementation there are still lots of things to follow up here:

- We don't play the correct entity/anim for "drowning" in lava.
- There's a somewhat unrelated bug if you change behaviour when falling which needs looking at.
- Performance wise this isn't great because we're saving the state every frame which involves serialising JSON which is slow. As an initial optimisation I want to only save every 250ms or something which should cut down the impact a lot. Even with the current implementation I don't see any noticeable hit to performance. 
- We should probably move away from the current approach of saving state when we have a more concrete design around saving games in general, saving and recalling just the position like this is a bit of a hack.
- Need to tie this together with the clover system and trigger game over if you die.

That being said this is a working implementation with no issues as far as I can see and works for both islands and iso envs.

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -23,12 +23,17 @@ export const BehaviourMode = {
 export function updateHero(game, scene, hero, time) {
     if (hero.props.dirMode !== DirMode.MANUAL)
         return;
+
     const behaviour = game.getState().hero.behaviour;
     handleBehaviourChanges(hero, behaviour);
     if (game.controlsState.firstPerson) {
-        processFirstPersonsMovement(game.controlsState, scene, hero);
+        processFirstPersonsMovement(game, scene, hero);
     } else {
-        processActorMovement(game.controlsState, scene, hero, time, behaviour);
+        processActorMovement(game, scene, hero, time, behaviour);
+    }
+
+    if (validPosition(hero.props.runtimeFlags)) {
+        scene.savedState = game.getState().save(hero);
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/game/loop/hero.ts
```diff
@@ -23,23 +23,40 @@ export const BehaviourMode = {
...
     if (game.controlsState.firstPerson) {
-        processFirstPersonsMovement(game.controlsState, scene, hero);
+        processFirstPersonsMovement(game, scene, hero);
     } else {
-        processActorMovement(game.controlsState, scene, hero, time, behaviour);
+        processActorMovement(game, scene, hero, time, behaviour);
+    }
+
+    if (validPosition(hero.props.runtimeFlags)) {
+        scene.savedState = game.getState().save(hero);
```
xesf, line 36: I think for now this is fine since it's just in memory, but when we decide to save it to the localStorage this could be a problem.

src/game/loop/physicsIso.ts
```diff
@@ -52,6 +52,9 @@ export function processCollisions(grid, _scene, obj, time) {
                 if (newY - position.y < 0.12) {
                     position.y = newY;
                     isTouchingGround = true;
+                    if (column.groundType === GROUND_TYPES.WATER) {
+                        obj.props.runtimeFlags.isDrowning = true;
+                    }
```
xesf, line 57: I believe we have other group types in isometric that would drown Twinsen. I am not sure we need to keep different flags for the other types.

slwilliams: so I did some testing and e.g. in the franco gazogem factory that wattery /gas stuff was still considered water so worked ok. There's probably some lava types that need looking at but I'll do that with the overall lava followup PR.

src/game/loop/hero.ts
slwilliams: ack ok yea, I think we need to think about how we want to approach save games and state in general and then change this accordingly.

PR #311, opened 2020-07-05
0.3.1 patch behavirour and music

xesf:
Based on 0.3.0 tag

* Fixed Audio Source Exception
* Allow User to enable Audio when context is suspended by the browser
* Allow behaviour selection within Behaviour Menu


xesf:
Looks like it does not like the changes in the change log

PR #314, opened 2020-07-09
Implement proto/jetpack

slwilliams:
Only a partial implementation for ISO environments, that's coming a followup PR so this remains reasonably small.

Works in VR (is quite fun!) I've mapped protopack and jetpack to be two modes after the first normal four for now. We probably want to revisit this when we get to thinking about how the inventory management should work within VR (that's going to be interesting!).

Couple of points I'd appreciate looking at:

src/game/actors.ts:221 seems to be the fix for the "cannot read skeleton of null" bug we've been seeing for a while, I'm not sure why the captured variable ever becomes null, but it seems more correct to me to be using this.animState anyway?

src/game/loop/animAction.ts:38 I think this is the right thin to do. Without gating the starting of the audio sample we end up with 100s of jetpack sounds playing because we start one playing for each frame. I think it makes sense to only have one repeating sample playing at once, but let me know if not.

src/game/actors.ts:295 Without this change we remove Twinsen from the scene before we place a new one. This causes the camera to jump away because it no longer has anything to track. By adding the new actor in first and then removing the old one this fixes that. However, I'm not sure if this is safe? I haven't noticed any issues when playtesting with this change.


src/game/loop/hero.ts
```diff
@@ -20,12 +20,22 @@ export const BehaviourMode = {
...
 
+// Behaviour modes which don't work with the default bodyIndex. If we force
+// Twinsen into one of these mode when he doesn't have the tunic we need to swap
+// to the tunic body.
+const BehaviourRequiresTunic = {
+    4: true, // PROTOPACK
+    5: true, // ZOE
+    6: true, // HORN
+    8: true, // JETPACK
+};
```
xesf, line 31: Can we look at the game flags instead?

Game Flags (set if in inventory):

> 0x08: Tunic and Sendell's medallion
> 0x18: Proto-Pack
> 0x2E: Horn of the Blue Triton
> 0x62: Super Jet-Pack

slwilliams: Turns out we don't need this special case at all and the game just does the right thing.

PR #315, opened 2020-07-10
Fix bug when you get stuck if you change behaviour whilst falling

slwilliams:
This also makes the ctrl key reset any animation (except drowning) like in the original game. This means you can jump and cancel the jump mid air like you used to be able to do. This is arguably a bug in the original implementation but I feel it was used so much as a core gameplay mechanic for us not to include it in the remake :) 

There are probably a few additional cases where we shouldn't be cancelling the playing animation, I've added a TODO for this so we can add them as we implement them.

### DIFF
--- a/src/game/actors.ts
+++ b/src/game/actors.ts
@@ -282,6 +282,9 @@ export async function loadActor(
         },
 
         cancelAnims() {
+            // TODO(scottwilliams): There are likely other cases where we should
+            // also not cancel the animations e.g. when Twinsen is crawling. Add
+            // them here when we've implemented them.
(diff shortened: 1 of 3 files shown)


agrande:
Do you think canceling the jump in mid-air in the original was a bug? It always seemed to me that it was an intended gameplay mechanic.

slwilliams:
I don't know for certain but I do think that e.g. using it to negate the effects of being pushed back when you get attacked seems like it shouldn't really be a gameplay mechanic. Certainly if you were designing something today you wouldn't plan on having a mechanic which relied on a technical implementation detail around cancelling in progress animations.

Issue #316, opened 2020-07-11
Can't properly see the "Replace by 3D model" button on the Iso Layouts Editor


felipesanches:
The button does not fit the UI. We can barely see a tiny bit of its top portion, but none of the text label as it overflows the available window area.

I tried resizing the browser window, but still can't make it fit. I also tried entering full-screen also to no avail.

The only way to see a slightly larger portion of the button is by tweaking the browser zoom level, but that completely messes up the rest of the UI.

![Screenshot from 2020-07-10 21-24-39](https://user-images.githubusercontent.com/213676/87212361-74773680-c2f4-11ea-96df-e218980e56ed.png)

felipesanches:
![Screenshot from 2020-07-10 21-39-15](https://user-images.githubusercontent.com/213676/87212614-d84e2f00-c2f5-11ea-9baa-63708bb3e958.png)

xesf:
You should be able to resize the area with the mouse. The size will be persisted next time you open the same Editor Tool

felipesanches:
I just tried it, but it also does not work. When resizing, the button still does not fit the available area:

![Screenshot from 2020-07-11 12-39-25](https://user-images.githubusercontent.com/213676/87227823-b4313300-c373-11ea-82e9-36531694fc99.png)

agrande:
Hmm, weird, it looks much bigger than in Chrome... I guess there's no specific styling on those buttons and it's using the browser defaults. I'll take a look at it.

agrande:
Can you check if the changes I made fixed the issue for you?
https://autopush.lba2remake.net/

felipesanches:
It got just a bit better, but still not fitting.

![Screenshot from 2020-07-11 14-03-14](https://user-images.githubusercontent.com/213676/87229556-48ed5e00-c37f-11ea-8317-4dc60441b95e.png)


I looked at your commit and tweaked the values. Making this change, it now looks good:
```javascript
const infoButton = {
    margin: '1px',
    padding: '0em'
};
```

![Screenshot from 2020-07-11 14-00-59](https://user-images.githubusercontent.com/213676/87229540-23f8eb00-c37f-11ea-9113-5b9cc42577e3.png)

agrande:
<img width="1439" alt="Screenshot 2020-07-11 at 22 43 10" src="https://user-images.githubusercontent.com/1733751/87233474-eff5e800-c3c7-11ea-86da-1acb7a0d3f22.png">
[...]

PR #320, opened 2020-07-11
fix size of buttons on iso layout editor for Firefox

felipesanches:
(issue #316)

**Preview here:** https://pr-320.lba2remake.net

### DIFF
--- a/src/ui/editor/areas/layouts/LayoutsEditorContent.tsx
+++ b/src/ui/editor/areas/layouts/LayoutsEditorContent.tsx
@@ -98,8 +98,8 @@ const applyChangesStyle = {
 };
 
 const infoButton = {
-    margin: '2px',
-    padding: '0.2em 0.4em'
+    margin: '1px',
+    padding: '0em'
 };
 
 const mainInfoButton = {


agrande:
Screenshot 2020-07-11 at 22 37 44
This is how it looks on chrome with your changes, 0 padding looks kind of ugly. Not sure what your browser is doing but it's adding unspecified padding there.

agrande:
Screenshot 2020-07-11 at 22 43 10
And this is what I have on Firefox with your change.

agrande:
In your screenshot there's clearly some padding applied, which shouldn't be the case. I think your browser is misconfigured, and not respecting the web standards. I can't accept this PR as it would deteriorate the experience for everyone else but you.

felipesanches:
ok, let's try to figure out a better solution then...

agrande:
What do you suggest?

felipesanches:
I don't know... I'll have to experiment and see if someone else has already dealt with a similar discrepancy.

agrande:
It looks weird, I've never seen that type of issue. I'm a bit at a loss here.

agrande:
> I don't know... I'll have to experiment and see if someone else has already dealt with a similar discrepancy.

Any news on this?

agrande:
I'm closing this as it seems to be an issue that is very specific to your own environment and the proposed fix makes things worse for everyone else. I'm leaving the related issue (#316) opened so we can keep investigating the root cause and find a better solution.

felipesanches:
thanks!

PR #326, opened 2020-07-20
Implement jetpack in ISO envs

slwilliams:
There are quite a few additional bug fixes here:

- Preserve animState when loading a new scene, this ensures we continue to jetpack correctly into the protection spell cave for example.
- Fix bug with floor height calculation in ISO envs. Before I was using the just the distance to the ground, rather than potential boxes under Twinsen.
- Be slightly more lenient with the fall height in ISO envs when using the protopack this ensures you don't drown if you protopack off the brown tiles in the protection spell cave. I think we've got something slightly off with our ISO collision detection but I see there's a task for that so won't spend more time here.
- Preserve Twinsen's Y value when doing a scene transition, this ensures if we jump through a door we don't just snap Twinsen to the ground.
- Finally, don't allow Twinsen to climb ladders when he's using the jetpack.

Still to fix:

- If you scene transition using the jetpack and then drown we haven't saved a most recent valid position so you get stuck in a drown loop, this should be straightforward to fix, but will do so in a follow up PR.

Let me know if I've missed something, probably have!

**Preview here:** https://pr-326.lba2remake.net

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -41,6 +41,9 @@ export function updateHero(game, scene, hero, time) {
 }
 
 function handleBehaviourChanges(scene, hero, behaviour) {
+    if (hero.props.runtimeFlags.isDrowning) {
+        return;
+    }
     if (hero.props.entityIndex !== behaviour) {
         hero.props.entityIndex = behaviour;
         hero.reloadModel(scene);
@@ -105,6 +108,7 @@ function processFirstPersonsMovement(game, scene, hero) {
                 game.getState().load(scene.savedState, hero);
                 hero.setAnim(AnimType.NONE);
                 hero.props.runtimeFlags.isDrowning = false;
+                hero.props.noInterpolateNext = true;
             });
             hero.animState.noInterpolate = true;
             return;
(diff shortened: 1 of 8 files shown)


xesf:
Found few things some related, some not:
* Jetpack animation seem to jump when animation loop starts. Very noticeable on ISO scenes.
* While drowning, if you keep changing behaviour, the animation will keep looping.
* Changing behaviour from ProtoPack to Jetpack under water works well, although the opposite no, Twinsen drowns. (I am not too sure this is actually the original behaviour but it could be right this way)
* Protopack to other Behaviour the transition is smooth. Jetpack transition to any other behaviour Twinsen falls to group. (this may expected as well)

slwilliams:
> Jetpack animation seem to jump when animation loop starts. Very noticeable on ISO scenes.

Yea this one I'm not sure about and looks to be tricky to solve so going to defer that to later.

>  While drowning, if you keep changing behaviour, the animation will keep looping.

Fixed this by both disallowing behaviour changes when Twinsen is drowning and also disabling interpolation when exiting the drowning state.

> Changing behaviour from ProtoPack to Jetpack under water works well, although the opposite no, Twinsen drowns. (I am not too sure this is actually the original behaviour but it could be right this way)
> Protopack to other Behaviour the transition is smooth. Jetpack transition to any other behaviour Twinsen falls to group. (this may expected as well)

These two are both intended (original behaviour).

PR #340, opened 2020-08-09
Bugfix/audio

xesf:
Audio fixes for both VR and non-VR modes.
Voice audio has been fixed to automatically skip to next voice in VR.

**Preview here:** https://pr-340.lba2remake.net


agrande:
Yay! 👍

PR #341, opened 2020-08-09
Fix actor rotation and a few other minor MOVE script bugs.

slwilliams:
The main issue was that we weren't turning in the shortest direction, I've changed it so that we look to see if it would be faster to turn anticlockwise or clockwise and then choose the shortest. There were a few other bugs being caused by the animation interpolation code we have so I've disabled interpolation for all script set_anim calls. This matches the original game (as far as I can tell by looking closely at the animations and how they switch).

This fixes the pharmacy issue with the umbrella thief not running out of the door correctly as well as other issues where actors were getting stuck due to not rotating correctly. This hasn't fully fixed the pharmacy scene - there is more work needed here e.g. Miss Amiradoo currently can't run out of the door for some reason. but the big issue is that we don't pause the game whilst text messages are printed on the screen. The original freezes the game and lets the user read the text. By not doing this and letting the move and life scripts progress things get out of sync and then things don't trigger correctly. This is something we'll need to resolve.

**Preview here:** https://pr-341.lba2remake.net

### DIFF
--- a/src/game/actors.ts
+++ b/src/game/actors.ts
@@ -150,14 +150,8 @@ export async function loadActor(
         goto(point) {
             this.physics.temp.destination = point;
             let destAngle = angleTo(this.physics.position, point);
-            const signCurr = this.physics.temp.destAngle > 0 ? 1 : -1;
-            const signTgt = destAngle > 0 ? 1 : -1;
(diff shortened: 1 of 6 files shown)


slwilliams:
Ah correction I haven't disabled animation interpolation, just overridden the frame we interpolate to, to be 0 rather than the loop frame.

PR #342, opened 2020-08-10
Behaviour menu model overlay

xesf:
This is just an initial PR, most likely won't be merge now, but just so you can check the code and play around once its deployed.

**Preview here:** https://pr-342.lba2remake.net


src/ui/game/BehaviourMenu.tsx
```diff
@@ -29,13 +39,189 @@ interface IBehaviourMenu {
...
+        skyColor: [0, 0, 0]
+    };
+    const ambience = {
+        lightingAlpha: 309,
+        lightingBeta: 2500
+    };
+    model[b] = await loadModel(
+        {},
+        b,
+        0,
```
slwilliams, line 198: You want to pass in the current bodyIndex here so that if Twinsen has his tunic you show the tunic model in the menu.

slwilliams (approved):
lgtm from a functionality point of view, I'm not as familiar with the rendering stuff here but code looks fine to me! ¯\_(ツ)_/¯

agrande (approved):

PR #343, opened 2020-08-11
Fix non-visible actors collision and distance calculation for GOTO points

slwilliams:
This fixes two bugs:

- We were performing collision detection for actors which weren't visible (i.e. having NO_BODY being called).
- We were calculating the distance to a MOVE script GOTO point using the centre of the actor not the edges of the bounding box. This meant that because MissAmiradoo has a bigger bounding box the distance to the centre of the actor was > 0.5 which is the distance we use to check to see if they've hit a flag.

We're now correctly setting the isVisible bool and checking that in the actor collision code and I've changed to calculate the distance by taking the min of each 4 points of the bounding box.

It's a bit weird we have actor.isVisible and actor.flags.isVisible, I've decided to use the former, but let me know if it's important we keep the flag.isVisible in sync too.

**Preview here:** https://pr-343.lba2remake.net

### DIFF
--- a/src/game/actors.ts
+++ b/src/game/actors.ts
@@ -3,7 +3,7 @@ import {cloneDeep} from 'lodash';
 
 import { loadModel, Model } from '../model';
 import { loadAnimState, resetAnimState } from '../model/animState';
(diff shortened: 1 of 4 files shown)


agrande:
Yay!

PR #345, opened 2020-08-11
Use meta (command) key instead of control on Mac OS for behaviour menu

agrande:
I tried preventing the default OS shortcuts (switching between desktops) from being triggered, using javascript, but it is not possible. Since most users will have those shortcuts enabled on Mac OS, I think it's better to move the behaviour key to another shortcut.

**Preview here:** https://pr-345.lba2remake.net


xesf:
Twinsen is now getting stuck turning left or right all the time when you use the Behaviour Menu.

agrande:
> Twinsen is now getting stuck turning left or right all the time when you use the Behaviour Menu.

@xesf That should be fixed now.

PR #348, opened 2020-08-15
Preserve bodyIndex on scene transition.

slwilliams:
Not sure when I broke this, but this fixes you losing your tunic when you exit the museum. 

**Preview here:** https://pr-348.lba2remake.net


agrande:
I think it was always broken.

PR #351, opened 2020-08-25
Initial implementation for the horn.

slwilliams:
For the moment the horn has unlimited hearts it can shoot, haven't worked on the sucking up mechanic.

Couple of other changes:

- I've called preventDefault for all key presses, this is because the alt key is used in Chrome as a shortcut with the arrow keys to navigate back and forth pages. Given we're going to wanting to combine alt a lot we need to stop this from happening. I think it's ok to do this this broadly, but let me know if not.
- I've changed the colours of the life and magic bars in the behaviour to match the original colours (used a colour picker to get the exact values).
- Allow other actors to collect hearts, but only hearts i.e. they can't collect keys etc.
- Life points in the game are actually scaled by 5x which is where that weird constant is coming from.


xesf (approved):
Did not test because it looks like PR has not been deployed.

All looks good, and its nice to see the previous extra and anim action implementation made it easier to integrate this. This seems promising to add the other anim actions including magic ball.

PR #352, opened 2020-08-26
Dome of the slate: VR remake (part 2)

agrande:
Sorry for the heavy PR, lots of stuff in here:

The `dome_env.ts` file is where most of the action happens, it handles all the dome specific shaders and effects.

- The stars. It's a custom shader that orients textured quads towards the camera, with some hack to half-lock them on the world z-axis to prevent it looking weird when tilting your head.
- The shooting stars (the sparkling trail solution is a bit hacky/glitchy, but hopefully pretty performant, I tested it with 100s of simultaneous shooting stars without any performance impact).
- A 3D model for the dome structure and additional geometry that wasn't in the original (made by @SendellX)
- The wall "window" effect using a custom shader applied to parts of the env 3D model (I call those windows "spots" in the code)
- A gradient shader on the entrance and exit platforms that transitions from the "LBA Shaders" used on the pillars and the wall effect shader.
- Some custom code to add stars on top of the corner pillars and making them (and their "cage") go up and down.
- A metadata system for replacing some sprite actors (the clover box and blue stars) with 3D models.
- Modifications to the ground/platforms disappearing effect to make the ground visible under those sprite actors (not too happy with my hard-coded solution there).
- Proper animation when dying in the stars in 3rd person (and some refactoring around that).
- A (dirty) hack to make you fall through the stars in 1st person.

**Preview here:** https://pr-352.lba2remake.net

If you want to try it without VR: https://pr-352.lba2remake.net/#editor=true&scene=26&iso3d=true, then press "C" to switch to free camera and look around.


xesf (approved):
What a huge PR :)

There is quite a lot of custom code that is only going to be used for the Dome, but I think that is inevitably with the overall change of gameplay and adaptability for VR.
Still, I see there are other things that will make the overall remake easier to extend like the sprites metadata.

I wonder if there is a way to keep VR as a separate add-on to the engine and not so embedded?! It may have the same challenges as adding LBA1 integration on the current engine.

This is really an amazing implementation and fun to play with. I still need to test the latest version with the enhancements but I appreciate the effort you both put into this. I guess its time to approach Fred and pitch for a official Vr experience of the LBA World.

src/game/loop/hero.ts
```diff
@@ -376,6 +349,50 @@ function processActorMovement(game, scene, hero, time, behaviour) {
     }
 }
 
+function checkDrowningAnim(game, scene, hero, time) {
+    if (hero.props.runtimeFlags.isDrowning
```
slwilliams, line 353: fwiw using early returns makes functions like these much easier to read:

```
function checkDrowningAnim(game, scene, hero, time) {
  if (!hero.props.runtimeFlags.isDrowning &&
      !hero.props.runtimeFlags.isDrowningLava
      !hero.props.runtimeFlags.isDrowningStars) {
    return false;
  }

  let anim = AnimType.DROWNING;
  ...
```

slwilliams (approved):
overall LGTM, there is a lot of complexity (and weird constants which I have no idea where they're coming from :D) but it's mostly contained within dome_env.ts which I think is ok.

Couple of trivial comments.

src/iso/metadata/models.ts
```diff
@@ -41,7 +41,7 @@ export async function loadFullSceneModel(entry: number, replacementData, isEdito
     : Promise<FullSceneModel> {
     const model = await loadModel(`/models/iso_scenes/${entry}.glb`);
     const threeObject = model.scene.children[0];
-    const heroPos = { value: new THREE.Vector3() };
+    const actorPos = times(5, () => new THREE.Vector3());
```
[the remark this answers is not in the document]
agrande: Well, that shader (dome_floor) is only used for the dome, but you’re right, it should be more generic if we want to use that type of floor somewhere else in the future (for some mods or something), I took a bit of a shortcut here. I’ll fix it when I can get my hands on my computer (I’m afk for the weekend).

src/game/loop/hero.ts
agrande: Absolutely, that’s what I usually do, I probably missed that opportunity here because I copy pasted that part from the call site(s) where it couldn’t use early return.

agrande: I’ll fix it before merging.

agrande:
Not sure how I can reduce the complexity. Maybe I can improve the naming a bit and replace magic numbers with named constants. Overall I’ve been working a bit in a rush on this (I have very little free time lately), and doing a lot of trial and error, leading to lower quality code. But since most of it is pretty much self-contained and unique to the dome (and probably won’t require much future maintenance), not sure it’s worth cleaning up too deeply. I’ll try to make it a bit more understandable before merging, though.

agrande:
As for keeping VR as a separate addon, I’m not entirely sure how best to do that. Some stuff are pretty intertwined with the original. Maybe a fork would work better for that? But I kind of imagine most of the VR stuff as not really specific to VR (the iso3d stuff for example) but more like building blocks for future general improvements and features that would be available to mod creators.

xesf:
> Not sure how I can reduce the complexity. Maybe I can improve the naming a bit and replace magic numbers with named constants. Overall I’ve been working a bit in a rush on this (I have very little free time lately), and doing a lot of trial and error, leading to lower quality code. But since most of it is pretty much self-contained and unique to the dome (and probably won’t require much future maintenance), not sure it’s worth cleaning up too deeply. I’ll try to make it a bit more understandable before merging, though.

You should go ahead and merge it.
No need to wait as it it self-contained and can be improved later if required.

xesf:
> As for keeping VR as a separate addon, I’m not entirely sure how best to do that. Some stuff are pretty intertwined with the original. Maybe a fork would work better for that? But I kind of imagine most of the VR stuff as not really specific to VR (the iso3d stuff for example) but more like building blocks for future general improvements and features that would be available to mod creators.

Not sure either, but its also great to be able to have everything contained in the same app, easy to deploy and easy to switch between versions. I think we should leave it for now and if we customise much more on VR maybe we can try to brainstorm something.

agrande:
Ok, I fixed the stuff that was outside dome_env.
I don't think I'll manage to take the time to clean up the dome_env code.
So I'm gonna merge this now.

PR #353, opened 2020-08-29
Fixed script issues with Twinsen being able to interact with dead actors

xesf:
**Preview here:** https://pr-353.lba2remake.net

### DIFF
--- a/src/game/scripting/life.ts
+++ b/src/game/scripting/life.ts
@@ -181,6 +181,8 @@ export function SUB_VAR_GAME(index, value) {
 }
 
 export function KILL_OBJ(actor) {
+    actor.props.life = 0;
+    actor.props.runtimeFlags.isDead = true;
     actor.isKilled = true;
     actor.isVisible = false;
     if (actor.threeObject) {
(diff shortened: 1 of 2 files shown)


src/game/scripting/life.ts
```diff
@@ -338,6 +342,7 @@ export function SUB_MAGIC_POINT(points) {
 }
 
 export function SET_LIFE_POINT_OBJ(actor, value) {
+    actor.props.runtimeFlags.isDead = false;
```
slwilliams, line 345: What if value is 0?

src/game/scripting/life.ts
```diff
@@ -583,6 +588,7 @@ export const SET_ARMOR_OBJ = unimplemented();
 export function ADD_LIFE_POINT_OBJ(index, points) {
     const actor = this.scene.actors[index];
     if (actor) {
+        actor.props.runtimeFlags.isDead = false;
```
slwilliams, line 591: and I guess here too in the unlikely case that props.life is zero and points is zero we shouldn't set isDead to false.

slwilliams (approved):

src/game/scripting/life.ts
xesf: Yeah you are right, we should prevent it as well, even though they use a separate command to kill objects.

src/game/scripting/life.ts
xesf: Indeed, will do this both fixes.
Cheers

PR #354, opened 2020-08-30
Initial implementation for combat.

slwilliams:
Still lots to do here namely:

- Need to check the angle of the actors so that you only hit if you're actually facing them.
- I've hacked the dying implementation for now, this isn't how it should be done and doesn't trigger things like bonuses or play the sample etc. need to work out how this should work.
- Need to ensure actor amour is taken into account when computing the amount of life to subtract. As it stands now Twinsen can kill any actor which shouldn't be the case.

**Preview here:** https://pr-354.lba2remake.net


src/game/actors.ts
```diff
@@ -301,7 +305,39 @@ export async function loadActor(
...
+            }
+
+            // TODO(scottwilliams): This doesn't take into account actor armour.
+            this.props.life -= hitStrength;
+            if (this.props.life <= 0) {
+                // TODO(scottwilliams): This isn't how this should be done, and
+                // doesn't trigger the correct things e.g. bonuses etc.
+                this.props.life = 0;
+                this.isKilled = true;
+                this.isVisible = false;
```
xesf, line 323: you're missing the runtimeFlag.isDead in there.
I guess isKilled is probably a duplication of this flag as well.

src/game/loop/hero.ts
```diff
@@ -20,6 +20,21 @@ export const BehaviourMode = {
...
 
+export const HandsHitLevels = [
+    5, // LVL_0
+    8, // LVL_1
+    18, // LVL_2
+    28, // LVL_3
+    38, // LVL_4
+];
+
+export const MagicBallHitLevels = [
```
xesf, line 31: Perhaps this should be part of the magicball object under game state file. So you can get the hit strength based on the current level.

src/game/loop/animAction.ts
```diff
@@ -16,7 +17,31 @@ export const ANIM = unimplemented();
...
+        if (a.index === hero.index) {
+            continue;
+        }
+        // TODO(scottwilliams): This doesn't take into account the actor angles.
+        if (distance2D(a.physics.position, hero.physics.position) < 1) {
+            const magicLevel = game.getState().hero.magicball.level;
+            a.hit(hero.index, HandsHitLevels[magicLevel]);
+        }
+    }
+};
```
xesf, line 44: These most likely can be used in the HIT scripting related commands.

xesf (approved):

src/game/actors.ts
slwilliams: ack added setting isDead.

src/game/loop/hero.ts
slwilliams: Moved this logic to state.ts

src/game/loop/animAction.ts
slwilliams: Not quite sure what you mean here, aren't these the HIT scripting commands already?

xesf: Nevermind, I didn't know we had already the Hit By Actor details in the scripts.
It seems there is only one SET_HIT_ZONE script to be implemented by I cannot remember what is that. Need to check on some old documentation.

xesf (approved):

src/game/loop/animAction.ts
slwilliams: ack sounds good.

PR #355, opened 2020-08-31
More complete combat implementation.

slwilliams:
- Actors now drop the correct bonus when they die and play the correct death audio sample.
- Use the correct hit strength when actors hit Twinsen, now he takes the correct amount of damage.
- Fix for runtimeFlags.isDead not being correctly set on actor revive.
- Use the correct life value for Twinsen when deciding if an actor is dead or not.

You can now almost fully play the Tralu scene except Ralph gets stuck running through the door due to our iso collision implementation and Twinsen gets stuck walking out of the scene with Zoe for some reason.

**Preview here:** https://pr-355.lba2remake.net

### DIFF
--- a/src/game/loop/animAction.ts
+++ b/src/game/loop/animAction.ts
@@ -3,7 +3,7 @@ import { each } from 'lodash';
 
 import { getRandom, distance2D } from '../../utils/lba';
 import { unimplemented } from '../scripting/utils';
-import { addExtra, ExtraFlag } from '../extras';
+import { addExtra, ExtraFlag, getBonus } from '../extras';
 import { SpriteType } from '../data/spriteType';
 
 export const NOP = unimplemented();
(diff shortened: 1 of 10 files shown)


src/game/loop/animAction.ts
```diff
@@ -16,29 +16,37 @@ export const ANIM = unimplemented();
 
 export const ANIP = unimplemented();
 
-export const HIT = (_action,  { actor, scene }) => {
+const DEATH_SAMPLE = 14;
+
+function processHit(actor, hitStrength, game, scene) {
     for (const a of scene.actors) {
-        if (a.index === actor.index) {
+        if (a.index === actor.index || a.isKilled || !a.isVisible) {
```
xesf, line 23: or runtimeFlags.isDead

xesf (approved):

agrande:
What's the difference between isDead and isKilled?

xesf:
> What's the difference between isDead and isKilled?

Was chatting with Scott the other day about it. I think we need to get rid of the isKilled flag which was added instead of the isDead.

src/game/loop/animAction.ts
slwilliams: ok I've removed all trace of isKilled in favour of runtimeFlags.isDead.

slwilliams:
Removed all trace of isKilled in favour of runtimeFlags.isDead.

agrande:
I think I actually introduced some new uses (At least 2) of isKilled in the dome PR.

agrande:
You may want to merge the latest origin/master into your branch to replace those too...

slwilliams:
oh ok, because I suck at Git I'll submit this as is and do an immediate follow up PR resolving the uses in the dome PR.

PR #360, opened 2020-09-02
Update changelog

xesf:


agrande:
There are new models for the tralü’s cave and school of magic too.

agrande:
Also, you can now select bricks in iso scène editor (and see their properties)

slwilliams:
Hey we actually got quite a lot of stuff done!

"Special implementation of the Dome of the Slate. You're going to be amused!!!" change that to "You're going to be amazed!!!"? amused is a bit of a weird word there.

other than that lgtm, lol @ you can punch FunFrock :D

slwilliams (approved):

agrande:
About punching FF, maybe we should actually punch him in the trailer! That would be cool!

slwilliams:
Yea should be doable! The reason I didn't for the video is because FunFrock isn't as easy a target as the casino guy and FunFrock is in an iso environment rather than an island which can look a bit weird in VR.

agrande:
Oh right, I didn’t think about the iso env part, and making him appear and
behave properly in another scene would be too much work.

xesf:
Thanks for the updates, I will do them later today.

xesf:
Changelog updated

CHANGELOG.md
```diff
@@ -20,13 +19,19 @@ _Unknown date_
 
 #### VR
 
-- Special implemtation of the Dome of the Slate. You're going to be amused!!!
 - Various audio improvements.
+- Special implemtation of the Dome of the Slate. You're going to be amazed!!!
```
agrande, line 23: didn't notice this, implemtation => implementation

xesf: Thanks for pointing that out.
I just pushed a change and will merge the current version.

PR #362, opened 2020-09-06
Found Inventory Objects Overlay

xesf:
**Preview here:** https://pr-362.lba2remake.net

### DIFF
--- a/src/model/inventory.ts
+++ b/src/model/inventory.ts
@@ -1,0 +1,102 @@
+import * as THREE from 'three';
+
+import { loadBody } from './body';
+import {
+    initSkeleton,
+    createSkeleton,
+    loadAnimState,
+} from './animState';
+import { loadMesh } from './geometries';
+import { loadTextureRGBA } from '../texture';
+import { createBoundingBox } from '../utils/rendering';
+import { loadLUTTexture } from '../utils/lut';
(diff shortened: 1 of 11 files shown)


slwilliams (approved):
lgtm, although I didn't test because looks like we didn't get a preview link.

src/model/inventory.ts
```diff
@@ -0,0 +1,102 @@
...
+        files,
+        bodies: [],
+        texture: loadTextureRGBA(files.ress.getEntry(6), palette),
+        mesh: null,
+        boundingBox: null,
+        boundingBoxDebugMesh: null,
+        materials: []
+    };
+
+    const body = loadBody(model, model.bodies, invIdx, null);
```
agrande, line 70: Oh, so you can actually call loadBody() directly, that's nice :)

agrande (approved):
That's a pretty big commit! Nicely done!
I like the generic overlay stuff.

agrande:
> lgtm, although I didn't test because looks like we didn't get a preview link.

Yeah, seems like I broken that when trying to fix the autopush stuff, ugh, my server code is too brittle.
I reloaded the server and the PR preview is now available (not for long though, I'll check it and merge if it works).

PR #363, opened 2020-09-13
Fix crash when rotating layout replacement model with "keepMat_" material

agrande:
**Preview here:** https://pr-363.lba2remake.net


slwilliams (approved):
lgtm

PR #369, opened 2020-10-02
Improve layouts variants matching

agrande:
I'm now searching for layout variants in several passes over the scene, starting with the biggest variants by volume, to the smallest, making sure I always match big layouts properly. I think I also simplified the code quite a bit overall.

This is potentially slower (although I didn't perceive much of a difference).
As the matching code also runs when just loading a scene in the game, it may make loading a bit slower. But I'll fix that soon by completely eliminating the need for the layout matching algorithm to run at load time, which will further simplify the code, and avoid potential inconsistencies between matching results at compile time vs load time. But that's for another PR.

There are also some fixes to the grid editor (the X and Z axis of the information panel were reversed when selecting a brick), and it's now properly saving the current scene being loaded in the editor, and displaying the list of scenes in the browser.

**Preview here:** https://pr-369.lba2remake.net


agrande:
Hmm, @xesf, @slwilliams wait for merging, I think I introduced a crash in some layouts.

agrande:
(@xesf @slwilliams) Ok, the crash is fixed.
There are actually a few layouts that break with this change (which I'll look into later), but I think overall it is better.

slwilliams (approved):
lgtm although I don't know very much about this code in general. Tested a bunch of places and all seems to work well.

xesf (approved):

PR #376, opened 2020-10-04
Minor iso collision fix

slwilliams:
This fixes the weird jumping / stuttering we see if there are iso bricks above Twinsen's head. This also allows us now to correctly walk up these stairs: https://imgur.com/a/pAKrVZw

I'm not 100% why this 2 * STEP delta is being used in the first place, as far as I can tell I don't break anything by doing this change and it seems to fix a bunch of problems so ¯\_(ツ)_/¯

Also fix a subtle bug where the hero actor angle is being set negative (we should probably be a bit clearer about trying to standardise on either 0-2Pi or -Pi to +Pi). This breaks the facing the camera logic when Twinsen is facing to the left.

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -359,6 +359,9 @@ function processActorMovement(game, scene, hero, time, behaviour) {
             if (!controlsState.sideStep) {
                 const euler = new THREE.Euler();
                 euler.setFromQuaternion(hero.physics.orientation, 'YXZ');
+                if (euler.y < -1) {
+                    euler.y += 2 * Math.PI;
+                }
                 hero.physics.temp.angle = euler.y;
                 if (controlsState.controlVector.y === 0) {
                     animIndex = controlsState.controlVector.x === 1
--- a/src/game/loop/physicsIso.ts
+++ b/src/game/loop/physicsIso.ts
@@ -61,7 +61,7 @@ export function processCollisions(grid, scene, obj, time) {
             if (obj.animState) { // if it's an actor
                 obj.animState.floorSound = column.sound;
             }
-            const minY = i > 0 ? bb.min.y - (2 * STEP) : -Infinity;
+            const minY = i > 0 ? bb.min.y : -Infinity;
             if (basePos.y >= minY) {
                 groundHeight = y;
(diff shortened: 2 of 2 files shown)


xesf:
I can't remember why we are using that step there, most likely something we have since the beginning.
In terms of angles, they were always a pain due to the various conversions, so it's possible issues are still there.

xesf (approved):

src/game/loop/hero.ts
```diff
@@ -359,6 +359,9 @@ function processActorMovement(game, scene, hero, time, behaviour) {
             if (!controlsState.sideStep) {
                 const euler = new THREE.Euler();
                 euler.setFromQuaternion(hero.physics.orientation, 'YXZ');
+                if (euler.y < -1) {
```
agrande, line 362: Where is this -1 coming from? It doesn't seem like a legit angle value, not sure I understand this...

src/game/loop/physicsIso.ts
```diff
@@ -61,7 +61,7 @@ export function processCollisions(grid, scene, obj, time) {
             if (obj.animState) { // if it's an actor
                 obj.animState.floorSound = column.sound;
             }
-            const minY = i > 0 ? bb.min.y - (2 * STEP) : -Infinity;
+            const minY = i > 0 ? bb.min.y : -Infinity;
```
agrande, line 64: I introduced the STEP in that commit: https://github.com/agrande/lba2remake/commit/54c0c8431696c7f1269734dfe07caabb38067c63.
If I recall correctly, it was to avoid the bug where Twinsen would fall through the floor (most likely when dropping frames). But I'm not sure at all anymore. Let's keep an eye on that bug, though...

slwilliams: ack ok I did test quite a bit and didn't observe any issues but ack will keep an eye out for it.

src/game/loop/hero.ts
slwilliams: oh crap yea you're right this should be "is euler.y negative" i.e. < 0 not -1. let me fix that. Surprisingly my test case still passed likely because the angle at that point would have happened to have been < -1...

PR #377, opened 2020-10-04
Load the sound type for objects and fix getGround to return it.

slwilliams:
I worked out that the scene object info contains the soundType int. Load that and ensure getGround correctly returns the object if Twinsen is standing on it. This fixes the crash we see if you were to walk on this pier for example: 

image

This includes a slight refactoring to abstract the object bounding boxes into objectInfo and bundle both the bounding box and other object info (including the soundType).


xesf (approved):
Very nice find.
PR deployment does not seem to be working, but everything looks ok.

agrande (approved):

PR #384, opened 2020-10-06
feat resource parsing by type

xesf:
This is more for letting you guys know the type of changes I am doing to avoid touching on some of this files.

I have introduced a resource type which matches the LBA extensions: http://lbafileinfo.kazekr.net/index.php?title=File_formats_by_extension
The idea is for each type we have a parsing function that will allow us to parse and create our own formats.

Only few of the things have changed and so far seems to be working fine. Had some weird issues with some scenes failing on getFloorHeight for some unknown reasons but that seems to be fine now.

The PR will also help us test a bit more this changes since there are a lot of times that have been changed.

**Preview here:** https://pr-384.lba2remake.net


xesf:
@agrande @slwilliams please have a look at the resource management changes to facilitate the LBA1 support.
Not fully finished, but to ensure we do not diverge from master too much, so it's better to merge it.

slwilliams (approved):
looks a reasonable refactor to me, tested a bit and doesn't look to have broken anything I can see.

src/scene/index.ts
```diff
@@ -1,407 +1,9 @@
...
-        isVisible: bits(staticFlags, 9, 1) === 0,
-        isSprite: bits(staticFlags, 10, 1) === 1,
-        isBackgrounded: bits(staticFlags, 13, 1) === 1,
-
-        noShadow: bits(staticFlags, 12, 1) === 1,
-        noElectricShock: bits(staticFlags, 17, 1) === 1,
-        noPreClipping: bits(staticFlags, 19, 1) === 1,
-    };
+export async function loadSceneData(index) {
+    return await getScene(index);
```
agrande, line 6: Why not calling getScene() directly here, instead of keeping the loadSceneData() function?

agrande (approved):
Looks good to me, tested too, with edge cases in the editor, doesn't seem to break anything.

src/scene/index.ts
xesf: The loadSceneData was the previous parsing function, but it looks like I forgot to clean up.

PR #386, opened 2020-10-07
Script and gameplay editor improvements

agrande:
- Automatic compilation of scripts on changes (instead of pressing a compile button)
- Various bits for making scripts more stable when compiling incomplete commands
- Adding new points and actors

**Preview here:** https://pr-386.lba2remake.net

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ _Unknown date_
 
 #### Editor
 
+- Gameplay editor: create new actors and points in scenes.
+- Scripts editor: now automatically compiles scripts on changes.
 - Iso editor bricks selection and properties.
 
 #### Bug fixes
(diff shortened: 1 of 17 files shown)


slwilliams:
hmm preview is failing to load with:

TypeError: Cannot destructure property 'objectToAdd' of 'this.props.sharedState' as it is undefined.

xesf (approved):

xesf:
Getting the same issue, but only when I am not running Editor mode. If you enable Editor mode it seems fine.

agrande:
Oh, right, makes sense, the GameUI component doesn’t have that prop when not in editor mode. I should have tested that. Thanks for noticing!

agrande:
I’ll fix it in a moment.

PR #397, opened 2020-10-13
changed rotation forward speed

jrneko:
this makes the rotation a bit more like the original

**Preview here:** https://pr-397.lba2remake.net


agrande (approved):
Lgtm, tested it, and it feels way nicer.

PR #400, opened 2020-10-14
Editor's window manager rework

agrande:
Up until now, the react components for editor areas were rendered as a binary tree, with the consequence that when an area was added or removed somewhere, the tree structure would change, and react would lose track of which component is what.

That means, for example, that for the main view in the gameplay editor, the react component would be unmounted and remounted when closing a side panel, and thus the game, renderer, scene, etc would all be reloaded as a result. To prevent that from happening I was using a saveMainData() function, responsible for saving an restoring the state of components when they were recreated. At the cost of some unwanted extra complexity in those components.

An additional problem with rendering the editor areas as a binary tree is that each area is positioned relative to their parent, which means they can't easily be moved around, turned into a floating window, minimised, or swapped with other areas.

So, the goal of this PR is to give each area some global coordinates and display them as a flat list with a stable react key, so they don't get destroyed at all when moved around. This simplifies the areas component code, keeping the complexity within the Editor component. The data structure for storing areas is still a binary tree, only the rendering part is flattened.

I made some additional styling changes in the PR regarding HTML focus, I'm now indicating clearly which area is focused, in a clean way, as opposed to the confusing frames appearing (or not) before around focused html elements.

**Preview here:** https://pr-400.lba2remake.net

### DIFF
--- a/src/ui/Editor.tsx
+++ b/src/ui/Editor.tsx
@@ -1,26 +1,23 @@
 import * as React from 'react';
 import {extend, each, concat, mapValues, cloneDeep} from 'lodash';
 import Area, {AreaDefinition} from './editor/Area';
-import {fullscreen} from './styles';
 import NewArea from './editor/areas/utils/NewArea';
 import {Type, Orientation} from './editor/layout';
 import {findAreaContentById, findMainAreas} from './editor/areas';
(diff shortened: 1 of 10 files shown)


xesf (approved):
Manage to test a bit and everything seems to be working fine.

PR #401, opened 2020-10-14
discreet mode turns faster now

jrneko:
there is still a slight delay before the next turn begins, but it is much closer to the original now

**Preview here:** https://pr-401.lba2remake.net

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -370,8 +370,9 @@ function processActorMovement(game, scene, hero, time, behaviour) {
                         : AnimType.LEFT;
                     let dy = 0;
                     if (hero.animState.keyframeLength) {
-                        const rotationSpeed =
-                        hero.props.entityIndex === BehaviourMode.DISCRETE ? 65 : 24;
+                        const rotationSpeed = hero.props.entityIndex === BehaviourMode.DISCRETE
+                         ? 65
+                         : 24;
                         const rotY = (hero.animState.rotation.y * rotationSpeed) / WORLD_SIZE;
                         dy = (rotY * time.delta * 1000) / hero.animState.keyframeLength;
                     }
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -3,7 +3,6 @@ import { DirMode } from '../actors';
 import { AnimType } from '../data/animType';
 import { WORLD_SIZE } from '../../utils/lba';
 import { processHit } from './animAction';
-import { ELSE } from '../scripting/structural';
 
 export const BehaviourMode = {
     NORMAL: 0,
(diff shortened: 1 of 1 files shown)


src/game/loop/hero.ts
```diff
@@ -370,8 +371,14 @@ function processActorMovement(game, scene, hero, time, behaviour) {
                         : AnimType.LEFT;
                     let dy = 0;
                     if (hero.animState.keyframeLength) {
-                        const rotY = (hero.animState.rotation.y * 24) / WORLD_SIZE;
-                        dy = (rotY * time.delta * 1000) / hero.animState.keyframeLength;
+                        if (hero.props.entityIndex === BehaviourMode.DISCRETE) {
+                            const rotY = (hero.animState.rotation.y * 65) / WORLD_SIZE;
```
agrande, line 375: Probably better to isolate the constant 65 | 24 into a variable like:
`const rotationSpeed = (hero.props.entityIndex === BehaviourMode.DISCRETE) ? 65 : 24`
to avoid duplicating the code.

agrande: Also, maybe this should make use of the speed property from the model.
@xesf do you know how to access that one?

agrande:
Also, the LGTM bot detected an issue: you added the line `import { ELSE } from '../scripting/structural';`, probably by accident (VSCode autocomplete I suspect).

jrneko:
> 
> 
> Hmm, better indent line 374 so it doesn't look like a separate instruction.
> Or better, if it fits, writing it as:
> 
> ```
> const rotationSpeed = hero.props.entityIndex === BehaviourMode.DISCRETE
>      ? 65
>      : 24;
> ```
> 
> I know I'm nitpicking a bit here, but having consistant style makes the code more readable overall.

sure! I'm here to learn

src/game/loop/hero.ts
```diff
@@ -370,7 +370,10 @@ function processActorMovement(game, scene, hero, time, behaviour) {
                         : AnimType.LEFT;
                     let dy = 0;
                     if (hero.animState.keyframeLength) {
-                        const rotY = (hero.animState.rotation.y * 24) / WORLD_SIZE;
+                        const rotationSpeed = hero.props.entityIndex === BehaviourMode.DISCRETE
+                         ? 65
```
agrande, line 374: Sorry for nitpicking again, but we usually indent with 4 spaces 😅

jrneko: ah lol

agrande:
No, I mean, like this:
```
const rotationSpeed = hero.props.entityIndex === BehaviourMode.DISCRETE
    ? 65
    : 24;
```

agrande:
Little explanation about the "why"...
The goal here is to be able to make the code easier to read quickly.
This is a single "instruction", but it's too big to fit in a single line (that would be too much info to process), so we split it on several lines, but we should be able to understand the structure of the code with just a glimpse.
Indentation helps with that, especially if we always stick to a consistent way of doing it.

agrande (approved):

PR #404, opened 2020-10-15
Loading optims

agrande:
A bunch of optimisations to improve loading times

**Preview here:** https://pr-404.lba2remake.net


slwilliams (approved):
nice this feels noticeably better.

PR #412, opened 2020-10-16
Initial art workflow documentation

Peterdevos:
You know what's more fun than modeling? Typing out documentation! (Just a starting point, this should hopefully come in handy in case more artists join)

**Preview here:** https://pr-412.lba2remake.net


doc/art_workflow.md
```diff
@@ -0,0 +1,86 @@
...
+The modelling software we use is [Blender](https://www.blender.org/download/). 
+
+In the current phase of the project, textures are seldom applied, so texturing software has low relevance.
+
+If you're using any kind of proprietary software, please make sure your copy has been properly licensed/purchased, and that the source/export files can also be opened/edited by open source alternatives to the used software.
+
+
+## Replacing 3D isometric layouts
+
+#### ![LayoutEditorFull](C:\Users\pvand\Desktop\LBA2Remake Workfiles\LayoutEditorFull.png)
```
agrande, line 20: You should replace this by a relative link, and have the picture in the same PR.

agrande (approved):
Great initiative!

doc/art_workflow.md
```diff
@@ -0,0 +1,86 @@
...
+For new 3D layout exports, the name should look something like this:
+
+`GB_082_088_wall_lamp_var02.glb`
+
+- The GB stands for Gray Buildings, which is the Layout Library the asset belongs to. (see 'Layout Library prefixes')
+- The number(s) that follow(s) represent which Layouts should use this asset. This can be multiple numbers, for instance if a layout used in different orientations.
+- The actual name of the object is up to us to define. Anything descriptive will do, as long as it's not overly lengthy. We stick to lowercase, and spaces are represented by underscores.
+- The 'Var02' represents a Layout Variant. These can be found in the editor under the Variants panel. Variants can for instance be a longer/shorter version of a Layout, or a part that has been split off. This suffix is used for actual Variant exports, starting from 'var01' (so, we do not name the main Layout 'var0' or anything like that).
+
+##### Layout Library prefixes
```
SendellX, line 55: I wrote something different in the PR where I changed the previous layout's names. DM for Dark Monk Statue, UN for Undergas (although UG is probably better) and CX for Island CX. I think I prefer having only two letters as a prefix, don't you think it's more consistent?

You're missing some libraries:
nº3: Caves (CA), nº12 Mosquibees (MO), nº14 Gazogem Refinery (GR) and nº15 Francos Island (FI).

doc/art_workflow.md
```diff
@@ -0,0 +1,86 @@
...
+
+##### Downloading an original 3D isometric layout as reference for creating the replacement
+
+These layouts can be downloaded from the Layout Editor by pressing "Download Layout" when selecting a Layout in the Iso Layout Editor. They will download as *.dae format files. 
+
+After importing them into Blender, they can serve as a basis/reference for modelling the replacing asset in terms of shape, size and dimensions. 
+
+Also note, that as the *.dae files do not self-contain textures, which will have to be downloaded by pressing the "Download Library Texture" button. This will download a shared texture atlas image which will correctly map to all Layouts that are downloaded from that Library.
+
+##### Creating the replacement 3D layout
```
SendellX, line 38: Maybe it would be nice to add that we're putting the .blend files into the AssetsSource folder, that makes the new layouts way easier to review.

doc/art_workflow.md
```diff
@@ -0,0 +1,86 @@
+# Art workflow
+
+These instructions describes the art workflow. It will explain how to properly create art assets, provide some technical details for how and why we do things, and describe best practices.
+
+## Artistic goal
+
+The goal is to adhere to the original art style of LBA as much as possible. Sometimes, liberties have to be taken in filling in previously unseen angles of assets. Also, due to the lower resolution of the original visuals, smaller details sometimes have to be interpreted by the artist. When we do, we try to respect the original art style as much as we can.
+
+## Software
```
SendellX, line 9: Should we add that the 2.82 version is the one that supports our .dae files (until the bug with the importer is fixed), and add a link to that specific version instead of the newest one?

doc/art_workflow.md
Peterdevos: I think sticking to 2 character prefixes is great! Using UG for Undergas does make more sense, so I'll keep that. I'll add the missing libraries to the list (it's normally unlike me to miss things like that... good catch).

doc/art_workflow.md
Peterdevos: I agree! I'll add a separate paragraph about our source files management. I can't find the folder for it yet (maybe it has not been created?), but we can discuss that together, it shouldn't block this initial document (which we'll be updating more in the future, I imagine).

doc/art_workflow.md
Peterdevos: I do dislike adding things in documentation that is prone to become outdated, but I can state that at the time of writing, we spotted this issue in Blender 2.83 which causes us to avoid that specific version. Would that be okay?

doc/art_workflow.md
Peterdevos: Ah, yes. My boo-boo. I'm now also noticing the /images subfolder in the docs folder. New to this... Haha....

doc/art_workflow.md
SendellX: I have the folder already, try fast-forwarding to the last master version maybe?

doc/art_workflow.md
SendellX: It doesn't work in 2.90 either. I'm downloading the newest version right now (I'll test it as soon as it finishes downloading) but I doubt it's fixed as I don't think anybody reported the bug. Maybe adding that if the importer gives an issue importing the collada layout they should try and use version 2.82 could be a solution? And putting a direct link to that webpage because I remember you had issues finding it.

doc/art_workflow.md
SendellX: I'm using SE for all the Sewers layouts, don't make me change all the names again please 😢 Also, do you have a specific reason why you prefer GA instead of GR for the Gazogem Refinery?

doc/art_workflow.md
```diff
@@ -0,0 +1,98 @@
...
+The modelling software we use is [Blender](https://www.blender.org/download/). (Note: at the time of writing, Blender v2.83 seems to have issues importing .dae files, so we're avoiding using that specific version.)
+
+In the current phase of the project, textures are seldom applied, so texturing software has low relevance.
+
+If you're using any kind of proprietary software, please make sure your copy has been properly licensed/purchased, and that the source/export files can also be opened/edited by open source alternatives to the used software.
+
+
+## Replacing 3D isometric layouts
+
+#### ![LayoutEditorFull](images\LayoutEditorFull.png)
```
agrande, line 22: I think you should put a / there rather than a \, the link doesn't seem to work.

doc/art_workflow.md
agrande: Maybe I can check for upgrading the exporter, or exporting as glb to fix the problem...

agrande: I don't remember why I picked dae instead of glb for the exporter to begin with, there must have been a reason.

doc/art_workflow.md
Peterdevos: No preference, and I think keeping SE so we avoid re-exporting a lot is a good reason. I'll change SW to SE, and change GA to GR.

doc/art_workflow.md
Peterdevos: Good to know... Frustratingly enough, it did work in Typora (which is the editor I use). I'll change it to a /.

agrande: Well, `\` is the path separator used on Windows, so I suppose that's the reason it works for you there, but on servers (usually Linux) and all other OSes `/` is the only correct path separator.

Peterdevos: I see! I learned something today :)

doc/art_workflow.md
Peterdevos: After fighting some more with my Git client, I managed to properly get to the latest master version and it's showing up!

doc/art_workflow.md
agrande: Seems to work now, not in the github diff view for some reason, but checking this file on your repo works.

agrande:
Maybe you should mention this file in the README, or in the getting_started.md file.

doc/art_workflow.md
SendellX: I just tried the newest Blender version and it's indeed still broken. Ugh 😞

PR #415, opened 2020-10-16
Refactor scene manager, renderer, and GameUI components

agrande:
First step towards converting the main game elements to JS classes, and simplifying the code.


xesf (approved):
Nice one, it is definitely better.

I made a quick fix, there was a missing binding causing the game to crash on Ask Choices.

agrande:
Thanks for the fix!
Next steps: Turning the scene object into a class and simplifying it.
Turning actors, zones, points, models, etc into classes too.

PR #416, opened 2020-10-17
Initial LBA1 Support (Batch 1)

xesf:
Not fully working yet, but a lot of the resource are already been parsed properly.

To test it you need to set the game param as "/#game=lba1" and have the LBA resources in the data folder.


agrande (approved):
lgtm, couldn't test as the import is not working yet on mac, but it doesn't seem to break anything.

Issue #179, opened 2018-03-27
Try to make 404 errors more explicit in dev-server logs.


agrande:
This would avoid confusion for new comers.

agrande:
Also, make them less verbose so they don't pollute the console.

Issue #149, opened 2018-03-04
Get loading image from game assets instead of having it in the repo


agrande:
![screenshot at 2018-03-03 09 33 14](https://user-images.githubusercontent.com/213676/36934304-f173bca6-1ec5-11e8-8304-2e79ea9d511c.png)

xesf:
The load imaged is an upscaled image using waifu2x.
Menu image has been enhaced as well and mirrored to have a wider image.

xesf:
Same issue applies to the LBA2 ribbon added for demo purposes.

PR #420, opened 2020-10-18
Fixed Distance check issues causing some actors to get Stuck. Luc/Tim…

xesf:
Fixed Distance check issues causing some actors to get Stuck. Luc/Timon Chez Luc, Racer in Desert Island racetrack

**Preview here:** https://pr-420.lba2remake.net

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -372,7 +372,7 @@ function processActorMovement(game, scene, hero, time, behaviour) {
                     if (hero.animState.keyframeLength) {
                         const rotationSpeed = hero.props.entityIndex === BehaviourMode.DISCRETE
                             ? 65
-                            : 24;
+                            : hero.props.speed;
                         const rotY = (hero.animState.rotation.y * rotationSpeed) / WORLD_SIZE;
                         dy = (rotY * time.delta * 1000) / hero.animState.keyframeLength;
                     }
--- a/src/game/scripting/move.ts
+++ b/src/game/scripting/move.ts
@@ -12,7 +12,7 @@ export function GOTO_POINT(point) {
         return;
     }
     const distance = this.actor.goto(point.physics.position);
-    if (distance > 0.65) {
+    if (distance > 0.55) {
         this.state.reentryOffset = this.state.offset;
         this.state.continue = false;
     } else {
--- a/src/game/scripting/move.ts
+++ b/src/game/scripting/move.ts
@@ -12,7 +12,7 @@ export function GOTO_POINT(point) {
         return;
     }
     const distance = this.actor.goto(point.physics.position);
-    if (distance > 0.55) {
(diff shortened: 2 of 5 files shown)


slwilliams:
Unfortunately this breaks a whole bunch of other stuff so it's not this simple :(

I'm not sure how best to approach this, we need to perhaps sit and think a bit about how this works in the original because I think something fundamental different is going on that what we're doing.

xesf:
Ahh, damn it. Didn't notice it broke other stuff :/. Do you have some examples I can look at?

This could be caused by the wrong speed rotation the actors have, not making them reach the point on time. Maybe something worth trying.

agrande:
Yeah, I think there’s more to it than the distance threshold. We looked at this in the past already. For Luc, I think the issue is that some invisible bricks under the countertop have collisions enabled while they don’t in the original. Not sure about the other cases.

agrande:
I wish we had an easy way to have automated tests for this, so we can make changes with having confidence of not breaking things.

xesf:
Yeah, not sure how we can achieve the testing here.

I believe for the invisible brick collision, that could be easily fixed by skipping those?

agrande:
> I believe for the invisible brick collision, that could be easily fixed by skipping those?

Yeah, but we have to learn how to identify those at load time.

xesf:
> > I believe for the invisible brick collision, that could be easily fixed by skipping those?
> 
> Yeah, but we have to learn how to identify those at load time.

Looks like the first entry of the LBA_bkg.hqr has that info.

This was taken from Zink's documentation:

First entry of lba_bkg file contains numbers from which the entries groups start and other information:

 Offset Type  Value  Meaning
              (dec)
  0x00  WORD    1    It means that entry #1 (counting from 0) is the first Grid
                      in the lba_bkg file.
  0x02  WORD   149   It means that entry #149 is the first Grid Fragment file in 
                      the lba_bkg file.
  0x04  WORD   179   It means that entry #179 is the first Library in the file.
  0x06  WORD   197   It means that entry #197 is the first Brick in the file.
  0x08  WORD  17903  It means that there is 17903 Bricks in the file.
  **0x0A  WORD   126   It means that Brick 126 (counting from 0) is transparent (used
                      for invisible walls) and don't have to be drawn.** 
  0x0C  DWORD 27151  It is the size in bytes of the largest Grid file (*.gr2)
                      decreased by 2. This value is used for memory management.
  0x10  DWORD 14164  It is the size in bytes of the largest Library file (*.bl2).
                      This value is probably used for memory management, but the
                      game won't crash (under Win XP) if the value is any other
                      (even zero).
  0x14  DWORD 387434 Unknown
  0x18  DWORD 86875  Unknown. This value is somehow connected with the largest amount
                      (or size?) of Layouts used in a Grid at once.

slwilliams (approved):

PR #421, opened 2020-10-19
Refacto scenes

agrande:
The main thing here is turning the Scene pseudo-class into a real JS class, and simplifying some stuff on the way to make things more readable and easy to navigate around.

**Preview here:** https://pr-421.lba2remake.net


slwilliams (approved):
refactor seems fine, tested a bunch of locations looks good to me.

PR #424, opened 2020-10-20
Scenery refacto

agrande:
Not complete yet, but better merge some stuff before it gets too complex.
Still some work to do on isometric sceneries to add more classes / type annotations.

**Preview here:** https://pr-424.lba2remake.net


slwilliams (approved):
this is a pretty big PR but the refactor seems to look ok to me. I've tested a bunch of cases and things seem to all still work.

agrande:
Yeah, I've tested this thoroughly, didn't see any regression, but there could be some.
I prefer to do the big refactos early in the release cycle for this reason.
Hopefully with all the new type annotations, future refactos should be safer too.

agrande:
I'll proceed with the merge then.

PR #425, opened 2020-10-21
Initial skeleton inventory

slwilliams:
Not functional, nor has any 3D model functionality yet. Just the initial structure for opening, closing and selecting a slot.

**Preview here:** https://pr-425.lba2remake.net

### DIFF
--- a/src/ui/GameUI.tsx
+++ b/src/ui/GameUI.tsx
@@ -102,7 +102,7 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
                 uiState.text === null &&
                 uiState.foundObject === null &&
                 !(uiState.showMenu || uiState.inGameMenu) &&
-                !this.props.uiState.inventory;
+                !uiState.inventory;
             if (showBehaviourMenu && this.isBehaviourKey(key)) {
                 this.props.setUiState({ behaviourMenu: true });
                 const scene = sceneManager.getScene();
@@ -117,7 +117,7 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
                 uiState.text === null &&
                 uiState.foundObject === null &&
                 !(uiState.showMenu || uiState.inGameMenu) &&
-                !this.props.uiState.behaviourMenu;
+                !uiState.behaviourMenu;
             if (showInventory && this.isInventoryKey(key)) {
                 this.props.setUiState({ inventory: !this.props.uiState.inventory });
                 if (game.isPaused()) {
--- a/src/ui/GameUI.tsx
+++ b/src/ui/GameUI.tsx
@@ -135,14 +135,14 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
             game,
         } = this.props;
 
-        if (this.isBehaviourKey(key)) {
+        if (this.props.uiState.behaviourMenu && this.isBehaviourKey(key)) {
             this.props.setUiState({ behaviourMenu: false });
             game.resume(false);
         }
     }
 
     isInventoryKey(key) {
-        return key === 'ShiftLeft';
+        return key === 'ShiftLeft' || key === 'ShiftRight';
     }
 
     isBehaviourKey(key) {
--- a/src/ui/GameUI.tsx
+++ b/src/ui/GameUI.tsx
@@ -15,6 +15,7 @@ import Ribbon from './game/Ribbon';
 import {KeyHelpIcon, KeyHelpScreen} from './game/KeyboardHelp';
 import { getVideoPath } from '../resources';
 import BehaviourMenu from './game/BehaviourMenu';
+import Inventory from './game/Inventory';
 import NoAudio from './game/NoAudio';
 import UIState from './UIState';
 import { SceneManager } from '../game/SceneManager';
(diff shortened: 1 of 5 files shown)


src/ui/GameUI.tsx
```diff
@@ -100,7 +101,8 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
                 uiState.ask.choices.length === 0 &&
                 uiState.text === null &&
                 uiState.foundObject === null &&
-                !(uiState.showMenu || uiState.inGameMenu);
+                !(uiState.showMenu || uiState.inGameMenu) &&
+                !this.props.uiState.inventory;
```
agrande, line 105: You can remove this.props here

src/ui/GameUI.tsx
```diff
@@ -109,27 +111,40 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
                 }
                 game.pause(false);
             }
+            const showInventory =
+                !uiState.loading &&
+                uiState.ask.choices.length === 0 &&
+                uiState.text === null &&
+                uiState.foundObject === null &&
+                !(uiState.showMenu || uiState.inGameMenu) &&
+                !this.props.uiState.behaviourMenu;
```
agrande, line 120: Same, this.props is not necessary.

src/ui/GameUI.tsx
```diff
@@ -109,27 +111,40 @@ export default class GameUI extends React.Component<GameUIProps, GameUIState> {
...
-            uiState,
             game,
         } = this.props;
-        const hideBehaviourMenu =
-            !uiState.loading &&
-            uiState.ask.choices.length === 0 &&
-            uiState.text === null &&
-            uiState.foundObject === null &&
-            !(uiState.showMenu || uiState.inGameMenu);
-        if (hideBehaviourMenu && this.isBehaviourKey(key)) {
```
agrande, line 127: Why did you remove this block?

src/ui/game/Inventory.tsx
```diff
@@ -0,0 +1,98 @@
...
+                break;
+            case 40:
+            case 'ArrowDown':
+                if (selectedSlot > inventoryColumns * inventoryRows - inventoryColumns) {
+                    newSlot = selectedSlot - inventoryColumns * (inventoryRows - 1);
+                } else {
+                    newSlot = selectedSlot + inventoryColumns;
+                }
+                setSelectedSlot(newSlot);
+                game.getState().hero.inventorySlot = newSlot;
```
agrande, line 57: This part is similar in all cases you can probably move it down to the if (action) part bellow.

src/ui/game/Inventory.tsx
```diff
@@ -0,0 +1,98 @@
...
+        };
+    });
+
+    const inventorySlots = [];
+    for (let i = 0; i < inventoryRows; i += 1) {
+        for (let j = 0; j < inventoryColumns; j += 1) {
+            const slot = i * inventoryColumns + j;
+            inventorySlots.push(
+                <div
+                  key={String(slot)}
```
agrande, line 80: Is it required to explicitly cast it as string?

agrande (commented):
Lgtm overall, just a few minor comments.

src/ui/GameUI.tsx
slwilliams: it's unnecessary, if the behaviour menu is already open non of these conditions can be true. There can't be any case where the menu is open and the user takes there finger off the ctrl key and it shouldn't close.

src/ui/game/Inventory.tsx
slwilliams: ah yea good point

src/ui/game/Inventory.tsx
slwilliams: apparently not, removed.

src/ui/game/Inventory.tsx
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    const inventorySlots = [];
+    for (let i = 0; i < inventoryRows; i += 1) {
+        for (let j = 0; j < inventoryColumns; j += 1) {
+            const slot = i * inventoryColumns + j;
+            inventorySlots.push(
+                <div
+                  key={slot}
+                  id={String(slot)}
```
agrande, line 75: Same question as with the key

slwilliams: oh right no id has to be a string.

agrande: Why do you need an id btw?

agrande:
After testing I found an issue when scrolling down with the down arrow key on the first column, it doesn't loop properly.
Also, the slots are rectangular when the aspect ratio is not the same as the original, but I suppose that can be improved in a future PR.

src/ui/game/Inventory.tsx
slwilliams: hmm yea that's a good point, removed for now.

slwilliams:
Fixed the boundary condition for the scrolling, yea the aspect ratio thing is interesting can take a look in a follow up PR.

agrande (approved):

PR #426, opened 2020-10-21
Make inventory square on all screen sizes

slwilliams:
**Preview here:** https://pr-426.lba2remake.net

### DIFF
--- a/src/ui/styles/inventory.scss
+++ b/src/ui/styles/inventory.scss
@@ -38,8 +38,8 @@
     user-select: none;
     list-style: none;
     font-size: 2em;
-    width: 98%;
-    height: 98%;
+    /*width: 115vh;
+    height: 98vh;*/
     position: absolute; 
     top: 50%;
     left: 50%;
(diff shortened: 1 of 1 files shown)


xesf (approved):

src/ui/styles/inventory.scss
```diff
@@ -49,6 +47,19 @@
     align-items: center;
 }
 
+@media screen and (orientation:landscape) { 
+    .inventory {
+        width: 115vh;
+        height: 98vh;
```
agrande, line 53: Nice thinking to use vh as a unit for this!

agrande (approved):

agrande:
If I'm not mistaken, the horizontal spacing between cells in the original is wider than the vertical.
Let me check this...

agrande:
Also, it doesn't work properly in the editor due to vh units being relative to the whole window.

agrande:
> If I'm not mistaken, the horizontal spacing between cells in the original is wider than the vertical.
> Let me check this...

My bad, it's the same spacing:
image
It's not properly aligned in the original, though! 😱

src/ui/styles/inventory.scss
slwilliams: yea I was glad this worked out well because otherwise it was going to be a little bit hacky.

slwilliams:
Ack on the editor not working bit I think that should be easy to fix but I've advanced the branch I'm using locally so I'll do this in a follow up PR.

PR #427, opened 2020-10-21
Adding Gamepad to support controllers in both mobile and desktop modes

xesf:
Support for gamepads. Not merging yet because I want to test this better on mobile, but I need a PR since I cannot test the controller while the phone is connect to the laptop. You can still review it though.

**Preview here:** https://pr-427.lba2remake.net

### DIFF
--- a/src/controls/GamepadControls.ts
+++ b/src/controls/GamepadControls.ts
@@ -49,23 +49,24 @@ export default class GamepadControls {
     dispose() {}
 
     update() {
-        const { game, state } = this.ctx;
-        if (!gamepadEventsFound) {
-            this.pollGamepads();
-        }
-
-        for (const index in this.controllers) {
-            const controller = this.controllers[index];
-            if (!controller.connected) {
-                continue;
+        const { state } = this.ctx;
+        const endTime = Date.now();
+        const elapsed = endTime - state.startTime;
+        if (elapsed > 100) {
+            state.startTime = Date.now();
+            if (!gamepadEventsFound) {
+                this.pollGamepads();
             }
 
-            this.resetButtonState();
+            for (const index in this.controllers) {
+                const controller = this.controllers[index];
+                // console.log(controller);
+                if (!controller.connected) {
+                    continue;
+                }
 
-            const endTime = Date.now();
-            const elapsed = endTime - state.startTime;
-            if (elapsed > 100) {
-                state.startTime = Date.now();
+                this.resetButtonState();
+
                 for (let i = 0; i < controller.buttons.length; i += 1) {
                     const button = controller.buttons[i];
                     const pressed = button.pressed || button.value > 0;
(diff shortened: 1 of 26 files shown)


src/controls/index.ts
```diff
@@ -6,6 +6,7 @@ import { VRControls } from './vr';
 import { SceneManager } from '../game/SceneManager';
 import Renderer from '../renderer';
 import Game from '../game/Game';
+import GamepadManager from './Gamepad';
```
agrande, line 9: Why use a different name for the file and class?
Why not calling both GamepadControls, to follow the same nomenclature as the rest?

src/controls/keyboard.ts
```diff
@@ -224,8 +225,7 @@ function focusOutHandler(game) {
     game.resetControlsState();
 }
 
-function resetCameraOrientation(game, sceneManager) {
-    const scene = sceneManager.getScene();
+export function resetCameraOrientation(game, scene) {
```
agrande, line 228: Can you add some type annotations on game and scene?
We'd better do this from now on, on known types.

agrande (commented):
You may want to try enabling controlsState.relativeToCam, it will probably feel more natural with a gamepad. I use this for VR.
Although it would require switching it on/off dynamically wether we used the keyboard or gamepad.

src/controls/index.ts
xesf: Yeah, need to change it. Initially was supposed to be just Gamepad but that collides with the native API.
I will rename it later.

src/controls/keyboard.ts
xesf: Sounds good, been doing this for missed that one.

xesf:
> You may want to try enabling controlsState.relativeToCam, it will probably feel more natural with a gamepad. I use this for VR.
> Although it would require switching it on/off dynamically wether we used the keyboard or gamepad.

Ok, will have a look to that one.
The gamepad fills a bit off and for some reasons when trying to test on mobile is not working, but locally works fine.

xesf:
@slwilliams @agrande can you have another look to the PR?
I've added loads of changes since the first version but everything should be there now except the Inventory navigation that can be added when Scott finishes the inventory piece.

There is one annoying bug though. I need to look further.
The left stick X axis, when the value is very very low, close to zero, but still positive, it acts as Negative.

agrande:
Lgtm, but I want to check it after I finish implementing what I'm doing on VR selector screen.

src/controls/GamepadControls.ts
```diff
@@ -0,0 +1,260 @@
+import Game from '../game/Game';
+import { SceneManager } from '../game/SceneManager';
+import { Params } from '../params';
+import { BehaviourMode } from '../game/loop/hero';
+import { resetCameraOrientation } from './keyboard';
+import { ControlsState, ControlActiveType } from '../game/ControlsState';
+// import { resetCameraOrientation } from './keyboard';
```
agrande, line 7: If you're not gonna use this in the end, maybe remove that line?

xesf:
Merged with master for the VR selection fixes

agrande (approved):
lgtm

PR #428, opened 2020-10-21
Initial functional implementation for inventory

slwilliams:
This is a functional implementation just with no 3D models or text yet. Items which are in the inventory show up green and you can press enter on them and they activate correctly. E.g. try giving the umbrella back to Leone.

**Preview here:** https://pr-428.lba2remake.net


src/ui/game/Inventory.tsx
```diff
@@ -1,11 +1,53 @@
...
 import '../styles/inventory.scss';
 
 const inventoryColumns = 7;
 const inventoryRows = 5;
 
-const Inventory = ({ game }: any) => {
+// inventoryMapping maps from inventory slot ID to item ID.
+// Note that slots are defined as starting from 0 in the top left and increasing
+// to the right row by row.
+const inventoryMapping = {
```
agrande, line 12: Maybe you can move this to a separate file (preferably just an external json, but just a separate .ts file should be enough to get started). This is so we can easily replace it for LBA1 or future mods.

src/ui/game/Inventory.tsx
```diff
@@ -49,13 +91,23 @@ const Inventory = ({ game }: any) => {
                 }
                 action = true;
                 break;
+            case 13:
+            case 'Enter':
+                const itemId = inventoryMapping[game.getState().hero.inventorySlot];
+                if (game.getState().flags.quest[itemId] === 1) {
+                    game.getState().hero.usingItemId = itemId;
```
agrande, line 98: This stays enabled forever, so if I select the umbrella in advance, for instance, and only then move to the see Leone (is that her name?) she automatically sees it without me going to the inventory.

xesf:
Btw, it will be great to skip the inventory menu with the ESC key as well.

The square inventory looks great.
Should we actually use the same width for the dialogues? It may look better if they share the same width

src/ui/game/Inventory.tsx
xesf: I second that. We can even add this as a Resource dependency so it will load a different mapping depending on the game.

src/ui/game/Inventory.tsx
xesf: Isn't this caused by the Script always checking the inventory for Leone?
Perhaps we need to add distance to Twinsen checks to skip those commands.

xesf (approved):

src/ui/game/Inventory.tsx
agrande: I don't think that's the case here, there is a distance check. What I did is select the umbrella in the inventory when still near the thief, then walked towards her, and as I arrived near her, without doing anything, she said thanks for the umbrella.

xesf: in that case we need to reset the _usingItemId_ at the end of the actors script, to guarantee that if none of the selected items have been used by one run, it won't be needed anymore.

agrande: Yeah, something like that.

src/ui/game/Inventory.tsx
slwilliams: I've refactored this stuff out and added a layer of indirection, let me know what you think.

slwilliams:
Also made esc close the inventory although it doesn't work the first time since the browser forces the escape key to be used to stop the app from grabbing focus, but nothing we can do about that.

src/game/data/inventory.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function GetInventoryRows() {
+    // TODO: Dynamically swap to use LBA once we have functionality for that.
+    return LBA2InventoryRows;
+}
+
+export function GetInventoryColumns() {
+    // TODO: Dynamically swap to use LBA once we have functionality for that.
+    return LBA2InventoryColumns;
+}
```
agrande, line 71: Maybe for a future PR, but I don't think you need function wrappers here, you can directly export the properties.

PR #430, opened 2020-10-22
Refactoring VR selector

agrande:
**Preview here:** https://pr-430.lba2remake.net

### DIFF
--- a/src/audio/index.ts
+++ b/src/audio/index.ts
@@ -33,6 +33,11 @@ export function createAudioManager(state) {
     voiceSource.setVolume(state.config.voiceVolume);
 
     return {
+        dispose: () => {
+            context.close();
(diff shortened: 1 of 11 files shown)


src/audio/index.ts
```diff
@@ -33,6 +33,11 @@ export function createAudioManager(state) {
     voiceSource.setVolume(state.config.voiceVolume);
 
     return {
+        dispose: () => {
+            context.close();
+            menuContext.close();
+        },
```
xesf, line 39: Nice one, totally forgot about it.

xesf (approved):

xesf:
Tested on mobile works as expected now.
Also tried VR on mobile and it works as well, touch works but with no controllers for now.

PR #431, opened 2020-10-23
Refactoring actors

agrande:
Another big refactoring around Actors.

Been testing this a lot, but it's possible that I could have broken some things, still, as there are some important changes in there.

This should significantly improve a few things:
- Navigation around the whole code base, as Actors are everywhere, and having type annotations makes it much easier to jump around the places where they are used
- Checking that we're not relying on some missing/renamed actor properties, as there's a whole bunch of them, and there were quite a few mistakes uncovered by the type annotations
- Helping us eliminate complexity around this central piece of the game engine

I renamed `runtimeFlags` to `state` so that things are more consistent with the rest of the codebase (adopting react naming conventions), and because it's less of a mouthful. I also removed a whole bunch of properties in there that were copy-pasted names from the original that we don't use at all in our engine. Overall, I think we should really name things our own way instead of sticking to the original names. And I think we should NEVER add variables/properties that we're not gonna use immediately (YAGNI), as it creates a lot of extra confusion as to what is what. Often we're doing things quite differently than the original, so adding the same flags as in there if we're not gonna end up using them is just polluting the code with useless complexity. And we can't know in advance what we're gonna use or not.

I also moved some stuff around between the actor object itself, and the props and state properties so that these things better reflect how we use them, and keep the class definition itself as simple as possible. I think we can go further on that path, especially with the actor props, I think these should be cleaned-up and should be exclusively treated as read-only data, which is not the case right now. I'll probably send another PR with such changes.

**Preview here:** https://pr-431.lba2remake.net


src/game/loop/hero.ts
```diff
@@ -194,7 +197,7 @@ function firstPersonPunching(game, scene) {
         const handPositions = game.controlsState.vrControllerPositions;
         for (let i = 0; i < handPositions.length; i += 1) {
             const intersect = ACTOR_BOX.containsPoint(handPositions[i]);
-            const velocity = game.controlsState.vrControllerVelocities[i];
+            const velocity = game.controlsState.vrControllerVelocities[i].length();
             if (intersect && velocity > PUNCH_VELOCITY_THRESHOLD && !punched[a.index][i]) {
```
agrande, line 201: @slwilliams you were comparing a THREE.Vector3 with a number here, I assume you meant to compare with the vector length? I wonder how the punching could work as I believe the comparison should have always returned false. Is there another place where you handle hits caused by punching?

slwilliams: that's weird vrControllerVelocities shouldn't be a Vector3 it's just a number. It's assigned just a number and that's why this code is treating it as one. I think I just got the incorrect initialisation in initializeVRController, that should just be = 0 not new THREE.Vector3().

agrande: oh, ok then, I'll change that

src/game/Scene.ts
```diff
@@ -154,6 +138,41 @@ export default class Scene {
...
+                    }
+                )
+            )
+        );
+        const zones = this.data.zones.map(props => loadZone(props, this.is3DCam));
+        const points = this.data.points.map(props => loadPoint(props));
+
+        this.actors.push(...actors);
+        this.zones.push(...zones);
+        this.points.push(...points);
```
xesf, line 160: Why not setting directly to the class entities above?

src/game/extras.ts
```diff
@@ -38,10 +40,17 @@ interface ExtraPhysics {
...
-
     flags: number;
     props: any;
+    state: {
+        isVisible: boolean;
+        isTouchingGround: boolean;
+        isDead: boolean;
+    };
+    threeObject?: THREE.Object3D;
+    sprite?: any;
```
xesf, line 53: Like mentioned above, a lot of those could shared with a Generic list of properties or a shared based class

src/ui/editor/areas/gameplay/scene/node_factories/variables.tsx
```diff
@@ -272,7 +272,7 @@ async function findAllRefsInSceneList(varDef, sceneList) {
 
 function findAllRefsInScene(varDef, scene) {
     const foundResults = [];
-    map(scene.actors, (actor, idx) => {
+    map(scene.actors, (actor, idx: number) => {
```
xesf, line 275: Missing type for Actor

src/ui/editor/labels.ts
```diff
@@ -15,7 +16,7 @@ const currentLabels = {
     zoneTypes: []
 };
 
-export function updateLabels(scene, labels) {
+export function updateLabels(scene: Scene, labels) {
```
xesf, line 19: Some of these we can also add the types: labels, selected, etc..

xesf (approved):
Wow, what a work!!!

I have pointed out various things but none really needs to be sorted as part of this refactoring, but more to point out for further discussions or PR.

Thanks for doing it!!!

src/game/Actor.ts
```diff
@@ -0,0 +1,536 @@
...
+        this.game = game;
+        this.scene = scene;
+        this.options = options;
+        this.props = cloneDeep(props);
+        this.physics = initPhysics(props);
+        this.state = Actor.createState();
+        this.state.isVisible = props.flags.isVisible
+            && (props.life > 0 || props.bodyIndex >= 0)
+            && props.index !== 1; // 1 is always Nitro-mecapingouin
+        this.isSprite = props.flags.isSprite;
```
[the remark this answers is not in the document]
xesf: I mean, the isSprite

src/game/Actor.ts
```diff
@@ -0,0 +1,536 @@
...
+interface ActorPhysics {
+    position: THREE.Vector3;
+    orientation: THREE.Quaternion;
+    temp: {
+        position: THREE.Vector3;
+        destination: THREE.Vector3;
+        angle: number;
+        destAngle: number;
+    };
+}
```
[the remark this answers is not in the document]
agrande: Yes indeed, I'm thinking of factoring this on a future PR.

src/game/Actor.ts
```diff
@@ -0,0 +1,536 @@
...
+    textColor: string;
+    prevEntityIndex?: number;
+    prevAnimIndex?: number;
+    prevAngle?: number;
+}
+
+interface ActorOptions {
+    isSideScene: boolean;
+    has3DCam: boolean;
+}
```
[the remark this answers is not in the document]
agrande: True, I should dig into how to avoid passing it to the actor, it's a bit tricky actually, I tried to get rid of that argument, but couldn't do it easily without breaking things.

src/game/Actor.ts
```diff
@@ -0,0 +1,536 @@
...
+    hasCollisionBricks: boolean;
+    hasCollisionBricksLow: boolean;
+    hasSpriteAnim3D: boolean;
+    isVisible: boolean;
+    isSprite: boolean;
+    canFall: boolean;
+    noShadow: boolean;
+}
+
+interface ActorProps {
```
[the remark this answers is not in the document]
agrande: Yeah, I'll rename it to Props for the scenes then. Makes sense.

src/game/Actor.ts
```diff
@@ -0,0 +1,536 @@
...
+            isSideScene: !scene.isActive,
+            has3DCam: scene.is3DCam
+        };
+        const props = createNewActorProps(scene, details);
+        const actor = await Actor.load(game, scene, props, options);
+        // This is usually called in 3 phases on all actors
+        // at scene load time. See loadScripts() function.
+        postProcessScripts(scene, actor);
+        cleanUpScripts(actor);
+        compileScripts(game, scene, actor);
```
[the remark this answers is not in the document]
agrande: Yeah, this method is used only in the editor for dynamically creating a new actor.
There's a comment on that above the function.

src/game/Scene.ts
```diff
@@ -53,7 +47,9 @@ export default class Scene {
     extras: any[];
     isActive: boolean;
     zoneState: {
-        listener?: Function;
+        skipListener?: Function;
+        currentChar?: number;
+        startTime?: number;
```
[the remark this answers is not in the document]
agrande: It's in the zones.ts file, but I have no idea what they are for.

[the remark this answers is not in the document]
agrande: Not even sure they are actually read, I think we're just writing to them.

src/game/Scene.ts
agrande: I could do that, I want the actors/zones/points arrays to be readonly so we can add remove some of them more dynamically (through the editor among other things, but also in code for multiplayer, for instance), without needing to worry about flushing the existing actors.
So I don't want to allow `this.actors = []` to happen anywhere but in the constructor, but maybe I should just do this instead, I was afraid it wasn't as readable, but it's probably not that bad:
```
this.actors.push(...await Promise.all<Actor>(
            this.data.actors.map(
                actor => Actor.load(
                    this.game,
                    this,
                    actor,
                    {
                        has3DCam: this.is3DCam,
                        isSideScene
                    }
                )
            )
        )
);
```

src/game/extras.ts
agrande: Not a base class, I think we should avoid using inheritance.
But a common interface yeah.

src/ui/editor/areas/gameplay/scene/node_factories/variables.tsx
agrande: That's because it's not the Actor class here, I got tripped up on this too. It's an ActorProps object, which I didn't add annotations for everywhere just yet, because this PR was already pretty big.

src/ui/editor/labels.ts
agrande: They are less critical, I think, the Actor, Scene, etc are critical because they're used all over the code base and contain a lot of complexity. Of course having more type annotations helps overall, but if we want to convert the WHOLE codebase, we're in for weeks of work. I prefer focusing on the parts that make a big difference for now.
The "low hanging fruits".

src/game/Actor.ts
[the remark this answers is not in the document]
agrande: Well, I don't think it belongs to state as it's not supposed to evolve over the lifecycle of the Actor, maybe it shouldn't be in the Actor class directly. It should stay only in the props, to keep only one source of truth.

src/game/Scene.ts
xesf: Ahh missed the readonly part, makes sense then, at least it is more readable than the version above.

agrande: You mean what I wrote in my comment is more readable, or what is in the code now?

xesf: Sorry, what is in the code now. Just better leave the way it it.

slwilliams:
fwiw I tested a bunch of the newer features I've done and everything seems to be fine

agrande:
Ok, I'm merging now, will address the remaining suggestions in the next PR.

src/game/Actor.ts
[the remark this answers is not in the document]
agrande: https://github.com/agrande/lba2remake/pull/440

src/game/Actor.ts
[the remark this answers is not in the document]
agrande: https://github.com/agrande/lba2remake/pull/441

PR #432, opened 2020-10-24
Lba1 grids

xesf:
Since not everything has been parsed, this won't work out of the box. You can still make it work, if you leave some of the not complete parsing function set has LBA2.

These are the resources that need to be change in order to make it work.

{ "strategy": "STATIC", "type": "ANM", "id": "ANIM",        "description": "Animations",            "path": "LBA2/ANIM.HQR" },
{ "strategy": "STATIC", "type": "LM2", "id": "BODY",        "description": "3d Models",             "path": "LBA2/BODY.HQR" },
{ "strategy": "STATIC", "type": "LS2", "id": "SCENE",       "description": "Scenes",                "path": "LBA2/SCENE.HQR" },
{ "strategy": "STATIC", "type": "LM2", "id": "OBJECTS",     "description": "Inventory Objects",     "path": "LBA2/OBJFIX.HQR" },
{ "strategy": "STATIC", "type": "TXR", "id": "BODY_TEXTURE",     "description": "3d Models Texture",        "path": "LBA2/RESS.HQR", "index": 6 },

**Preview here:** https://pr-432.lba2remake.net


src/game/scenery/isometric/IsoScenery.ts
```diff
@@ -40,11 +40,11 @@ export default class IsoScenery {
         const [palette, bricks, gridMetadata, mask] = await Promise.all([
             getPalette(),
             getBricks(),
-            getGridMetadata(sceneryIndex + 1),
+            getGridMetadata(sceneryIndex),
```
agrande, line 43: Nice change! Tested it in various situation and seems stable.

www/resources/lba1.json
```diff
@@ -28,6 +28,8 @@
         { "strategy": "STATIC", "type": "M4A", "id": "VOICES_GAM",       "description": "Main Voices",  "path": "LBA/VOX/%LANGVOICECODE%_GAM_AAC.VOX.zip" },
         { "strategy": "STATIC", "type": "M4A", "id": "VOICES_000",       "description": "Voices 0",     "path": "LBA/VOX/%LANGVOICECODE%_000_AAC.VOX.zip" },
 
+        { "strategy": "STATIC", "type": "JSON", "id": "MODEL_REPLACEMENTS", "description": "Models Replacement",    "path": "../../metadata/model_replacements.json" },
```
agrande, line 31: We'll probably want a separate file for LBA1, but it doesn't matter much at this point, I guess.

agrande (approved):
Lgtm

www/resources/lba1.json
xesf: Yeah, I think all the current metadata should be moved to a lba2 folder and be compatible for lba1 in case we want to start replacing those as well. Not important, but better do that change earlier so we don't add lba2 specific stuffs there.

PR #436, opened 2020-10-24
Improve layouts matching

agrande:
**Preview here:** https://pr-436.lba2remake.net


src/game/scenery/isometric/layouts.ts
```diff
@@ -0,0 +1,38 @@
+const cachedLayouts = {};
+
+export function loadLayout(bkg, layout) {
+    const { library, index } = layout;
+    const key = `${library}_${index}`;
+    if (key in cachedLayouts) {
+        return cachedLayouts[key];
+    }
+    const buffer = bkg.getEntry(179 + library);
```
xesf, line 9: Can we parse the Layout using the Resource parsing?
This will fail for LBA1 integration.

src/game/scenery/isometric/metadata/metadata.ts
```diff
@@ -1,7 +1,10 @@
 import { map } from 'lodash';
 import { loadModel } from './models';
+import { getBricksHQR } from '../../../../resources';
```
xesf, line 3: getBricksHQR or any Resource helper ending as HQR are temporary.
We should try to use getBricks/getLibraries/getGrids instead to avoid any custom parsing.

src/ui/editor/areas/isoGrid/browser/IsoBrowserArea.tsx
```diff
@@ -77,7 +77,7 @@ async function collectIsoScenes(location, scenes, push = true) {
                 const sceneMap = await getSceneMap();
                 const gridData = new DataView(
                     bkg.getEntry(
-                        sceneMap[newLocation.props[0].value].index + 1
+                        sceneMap[newLocation.props[0].value].sceneryIndex + 1
```
xesf, line 80: This will also fail for LBA1 integration.

xesf (approved):

src/game/scenery/isometric/layouts.ts
agrande: Sure, I'll do that tonight.
Not sure it will be enough for the editor to work with LBA1, there may be other things to adapt.

src/game/scenery/isometric/metadata/metadata.ts
agrande: Hmm, right, I'll see how I can adapt that code.

src/ui/editor/areas/isoGrid/browser/IsoBrowserArea.tsx
agrande: Why?

src/game/scenery/isometric/metadata/metadata.ts
xesf: I think that part is very custom, reason I did not change it in the first place.

src/ui/editor/areas/isoGrid/browser/IsoBrowserArea.tsx
xesf: The "+ 1".
For LBA1 the scenes and grids and layouts match their entries, while in LBA2 grids are 1 index ahead and layouts are reused indexes.
This one should also be extracted from the Grid resources to make sure the Grid and Library indexes match for both LBA1 and LBA2.

PR #438, opened 2020-10-25
Compile LBA1 Scripts

xesf:
**Preview here:** https://pr-438.lba2remake.net

### DIFF
--- a/src/game/scripting/data/lba1/condition.ts
+++ b/src/game/scripting/data/lba1/condition.ts
@@ -1,0 +1,195 @@
+import * as lc from '../../condition';
+import { unimplemented } from '../../utils';
+
+export const ConditionOpcode = [
+    {
+        opcode: 0x00,
+        command: 'COL',
+        handler: lc.COL,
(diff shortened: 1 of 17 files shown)


agrande (commented):
Are the opcodes so different between the 2 games as to require separate tables?
I though LBA2 scripts were retro-compatible with LBA1.

xesf:
> Are the opcodes so different between the 2 games as to require separate tables?
> I though LBA2 scripts were retro-compatible with LBA1.

Subtle changes, but not sure how we can reuse.
There are some Opcodes not required for LBA2 and they simply replaced by different meaning and opcode contract.
There are also opcodes with different bytes, like some in LBA2 required 2 bytes, while LBA1 just a single byte.

Also the list of opcodes is shorter on LBA1, but all opcodes with same name are reusing the same implementation.

agrande:
Ok, sgtm, let me test the script editor to see if everything is still working as expected.

agrande (approved):

PR #439, opened 2020-10-25
Fixed Text parsing for LBA1

xesf:
**Preview here:** https://pr-439.lba2remake.net


src/resources/parsers/text1.ts
```diff
@@ -0,0 +1,31 @@
...
+                charmap[data.getUint8(i)]
+                : data.getUint8(i))
+            );
+        }
+        texts[mapData[idx]] = {type, index: idx, value: value.join('')};
+        idx += 1;
+    } while (end < data.byteLength);
+
+    return texts;
+};
```
agrande, line 31: This looks pretty similar to me to the LBA2 parser, can't we factorise that somehow?

xesf: We can yes, very tiny changes. Let me try to do that now.

agrande (approved):
lgtm

PR #448, opened 2020-10-26
Zone refacto to class

agrande:
**Preview here:** https://pr-448.lba2remake.net


src/game/Zone.ts
```diff
@@ -0,0 +1,147 @@
...
+export interface ZoneProps {
+    sceneIndex: number;
+    index: number;
+    type: number;
+    pos: number[];
+    snap: number;
+    info0: number;
+    info1: number;
+    info2: number;
+    info3: number;
```
xesf, line 42: Any idea on how we can improve the props when they need to be called from outside Zones?

Turns out, the allocation of the info[X] and snap are different from LBA1 and LBA2.
They also have different meanings depending on the Zone Type, but we are using them for collision, scenary checks.

It will be great to have for example targetIndex instead of snap/info0, or x,y,z instead of info0, info1, info2 exposed to outside, but I don't see what is the best way to expose those in the Zone class.

xesf (approved):

src/game/Zone.ts
agrande: Well, I imagine we could use a union type for this, with different types for each zone.

xesf: Nice, completely forgot about the union types, it may work.

PR #460, opened 2020-10-31
Parsing Libraries properly for LBA1 and fixed floor samples

xesf:
**Preview here:** https://pr-460.lba2remake.net

### DIFF
--- a/src/game/loop/animAction.ts
+++ b/src/game/loop/animAction.ts
@@ -6,6 +6,7 @@ import { addExtra, ExtraFlag, getBonus } from '../extras';
 import { SpriteType } from '../data/spriteType';
 import { SampleType } from '../data/sampleType';
 import Actor from '../Actor';
+import { getParams } from '../../params';
 
 export const NOP = unimplemented();
 
@@ -85,9 +86,10 @@ export const SAMPLE_STOP = (action, { game }) => {
 export const ZV = unimplemented();
 
 export const LEFT_STEP = (_action, { game, scene, animState }) => {
+    const isLBA1 = getParams().game === 'lba1';
     const floorSound = animState.floorSound;
     if (floorSound !== undefined && floorSound !== -1) {
-        const offset = scene.props.isIsland ? 30 : 60;
+        const offset = scene.props.isIsland ? 30 : isLBA1 ? 126 : 60;
         const sampleIndex = floorSound + offset;
         // const frequency = getRandom(0, 0x1000) + 3596;
         const audio = game.getAudioManager();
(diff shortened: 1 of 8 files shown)


src/game/loop/animAction.ts
```diff
@@ -96,9 +98,13 @@ export const LEFT_STEP = (_action, { game, scene, animState }) => {
 };
 
 export const RIGHT_STEP = (_action, { game, scene, animState }) => {
-    const floorSound = animState.floorSound;
+    const isLBA1 = getParams().game === 'lba1';
+    let floorSound = animState.floorSound;
+    if (animState.floorSound2) {
```
slwilliams, line 103: hmm I'm a bit confused what floorSound2 represents. It is correct that only the right step should use floorSound2?

src/resources/parsers/libraries.ts
```diff
@@ -26,11 +26,14 @@ const parseLayout = (dataView: DataView, index: number) => {
...
         const type = dataView.getUint8(offset + (i * 4) + 1);
+        const shape = dataView.getUint8(offset + (i * 4));
+        const brick = dataView.getUint16(offset + (i * 4) + 2, true);
         blocks.push({
-            shape: dataView.getUint8(offset + (i * 4)),
+            shape,
             sound: bits(type, 0, 4),
-            groundType: bits(type, 4, 4),
-            brick: dataView.getUint16(offset + (i * 4) + 2, true)
+            sound2: isLBA1 ? bits(type, 4, 4) : null,
```
slwilliams, line 34: Could you add a comment here explaining what sound2 is?

src/resources/parsers/libraries.ts
```diff
@@ -26,11 +26,14 @@ const parseLayout = (dataView: DataView, index: number) => {
...
+            groundType: isLBA1 ? (type & 0xFF) === 0xF1 ? 1 : 0 : bits(type, 4, 4),
```
slwilliams, line 35: unrelated, but in LBA1 there's only 2 groundTypes? 1 or 0?

slwilliams (approved):

src/game/loop/animAction.ts
xesf: yeah, the idea is left uses sound1 and right uses sound2

src/resources/parsers/libraries.ts
xesf: yes, only water exist in LBA1.

src/resources/parsers/libraries.ts
xesf: Sure, just pushed the comments.

Perhaps we should start creating documentations for each of these file formats and link them to the parser.

PR #461, opened 2020-10-31
Fixed Jump while running for LBA1

xesf:
**Preview here:** https://pr-461.lba2remake.net


src/game/loop/hero.ts
```diff
@@ -314,9 +316,10 @@ function processActorMovement(
             }
         }
         if (controlsState.jump === 1) {
+            const isLBA1 = getParams().game === 'lba1';
```
slwilliams, line 319: Hoist this to the top of the function? I expect we'll certainly going to use this some more in this function. (and the same for above)

slwilliams (approved):

PR #464, opened 2020-10-31
Add functionality for pre and post game loop functions

slwilliams:
This means we only have the usingItemId set for a single game loop iteration. This functionality will likely be more generally useful too.

**Preview here:** https://pr-464.lba2remake.net


xesf (approved):
I like the idea and it's very versatile.

PR #466, opened 2020-11-02
Game loop refactoring

agrande:
This is not finished yet, although I want to make (relatively) small PRs, so I'm pushing it in the current state, as it's stable and can be merge if there's no regression.
I want to simplify the Actor class, and eliminate the remaining files in the loop folder by moving the code into some new classes that handle physics, player controls and animState in a more isolated and modular way.

**Preview here:** https://pr-466.lba2remake.net


src/game/Game.ts
```diff
@@ -183,4 +268,37 @@ export default class Game {
...
+                currentAmb &= 3;
+            }
+            const { sampleMinDelay, sampleMinDelayRnd } = scene.props.ambience;
+            scene.props.ambience.sampleElapsedTime =
+                time.elapsed + (getRandom(0, sampleMinDelayRnd) + sampleMinDelay);
+        }
+        if (scene.props.ambience.sampleMinDelay < 0) {
+            scene.props.ambience.sampleElapsedTime = time.elapsed + 200000;
+        }
+    }
```
xesf, line 303: This makes more sense to be part of the Scene class

xesf (approved):
Indeed better to do this type of refactoring in batches to be quickly reviewed and merged as they delete and change quite a few logics.

I will hold on the Audio changes for the positional audio as it touches in a lot of these files.

src/game/Game.ts
agrande: True, I wasn't sure of the impact of moving it over there. I suppose not much. I'll do that tonight.

PR #469, opened 2020-11-04
3D Audio

xesf:
Just wanted to push this so you can can test a bit.
Mainly only Actor related stuff is using the Positional Audio, but the whole idea of having the separate Audio Manager changed a bit.

You can have a look at how this is currently done, even though it needs some clear up. If you have other ideas, please let me know.

Possible Issue: I am seeing an increase of Listeners been added and looks like a leak.

https://pr-469.lba2remake.net/#audio3d=true (by default)
https://pr-469.lba2remake.net/#audio3d=false

**Preview here:** https://pr-469.lba2remake.net


src/audio/Audio.js
```diff
@@ -0,0 +1,391 @@
+// from ThreeJS: https://github.com/mrdoob/three.js/blob/master/src/audio/Audio.js
```
agrande, line 1: Why not use the original by importing it directly?

xesf: Some changes that need to be done, but that's not fully implemented yet.

The Listener is the main piece because it's creating a new audio context all the time, and we want to reuse the same. Had some issues with typing after that but still need to further look and see if I can use both PositionalAudio and Audio from ThreeJS but only implement the Listener in our code.

agrande: Why is ThreeJS doing it that way? Can't we make the change in ThreeJS if maintaining a single context is more efficient (is it?), and send them a PR?

xesf: We can indeed, and perhaps use patcher for the time being to fix that code straight in the node_modules.
Its a small change actually and there is value on doing it.

xesf: Looking at their code I found out there is a way to set the Context, so no need to create a PR or have that code on our side.

slwilliams:
had a play, all the audio now seems very weird, "underwater" like haven't looked at the code yet but are you doing some weird stuff with the cutoff frequency?

xesf:
> had a play, all the audio now seems very weird, "underwater" like haven't looked at the code yet but are you doing some weird stuff with the cutoff frequency?

Yeah, I added the same filter but got different results from some reason. Changed the frequency function but thought it was the same. I may be using the wrong Hz number to do the math.

Some sounds do not seem to be playing now, not sure because of the frequency change, I will test that.

The side scenes are having the wrong behaviour when changing scenes, the sounds stops and may be to do with the listener attached to the camera. The camera is the same for child scenes, but for some reason the listener does not seem to update properly.

src/audio/index.ts
```diff
@@ -75,9 +83,62 @@ export function createAudioManager(state) {
...
-            const sampleSource = createSampleSource(context);
-            sampleSource.setVolume(state.config.soundFxVolume);
+        createSamplePositionalAudio: (): THREE.PositionalAudio => {
+            const sound = new THREE.PositionalAudio(listener);
+            sound.setDirectionalCone(60, 90, 0.3);
+            sound.setRolloffFactor(40);
+            sound.setRefDistance(20);
+            sound.setMaxDistance(10000);
+            sound.setVolume(1);
+            sound.setFilter(sound.context.createBiquadFilter());
```
slwilliams, line 93: So the default filter for createBiquadFilter() is a low pass filter which is why everything sounds underwater. This isn't the way the frequency parameter is meant to be used. The current implementation uses an allpass filter which also isn't correct (the samples don't do the right thing when the frequency changes).

I think this needs more looking into but for now using a lowpass filter is much worse than what we have now. To be honest I would just remove these filters all together, they're not adding anything.

src/cameras/3d.ts
```diff
@@ -28,6 +29,12 @@ export function get3DCamera() {
     orientation.matrixAutoUpdate = false;
     controlNode.add(orientation);
     orientation.add(camera);
+
+    if (game) {
+        const audio = game.getAudioManager();
+        camera.add(audio.listener);
```
slwilliams, line 35: I think attaching the listener to the camera is the wrong thing to do. Wouldn't Twinsen be more appropriate? At the moment you can be in situations here you swing the camera around and Twinsen is in the same place but the audio varies widely and it seems a bit weird.

src/game/Extra.ts
```diff
@@ -92,15 +93,17 @@ export default class Extra {
...
         time: Time
     ): Promise<Extra> {
-        const extra = new Extra(position, angle, spriteIndex, bonus, time);
+        const extra = new Extra(game, position, angle, spriteIndex, bonus, time);
         extra.init(angle, 40, 15);
         await extra.loadMesh(scene);
         scene.addExtra(extra);
-        extra.playSoundFx(game, SAMPLE_BONUS);
+        const audio = game.getAudioManager();
+        audio.playSound(extra.sound, SAMPLE_BONUS);
```
slwilliams, line 101: This doesn't seem to work, both the sound effect for a bonus found and being picked up doesn't seem to work any more.

src/game/scenery/island/environment/Lightning.ts
```diff
@@ -114,13 +115,18 @@ export default class Lightning {
...
         : 381;
-        const sample = audio.playSample(index);
-        sample.setVolume(volume);
+
+        if (!this.sound) {
+            this.sound = audio.createSamplePositionalAudio();
+            this.sound.setDirectionalCone(360, 0, 0);
+            this.threeObject.add(this.sound);
+        }
+        audio.playSound(this.sound, index);
```
slwilliams, line 128: This doesn't seem to work because I can't hear the lightening anymore.

src/audio/index.ts
```diff
@@ -75,9 +83,62 @@ export function createAudioManager(state) {
         },
 
         // samples
-        playSample: (index: number, frequency: number = 0x1000, loopCount: number = 0) => {
-            const sampleSource = createSampleSource(context);
-            sampleSource.setVolume(state.config.soundFxVolume);
+        createSamplePositionalAudio: (): THREE.PositionalAudio => {
+            const sound = new THREE.PositionalAudio(listener);
+            sound.setDirectionalCone(60, 90, 0.3);
```
slwilliams, line 88: I don't think we want a directional cone here, just an omnidirectional sound would be better in my opinion, why did you choose a direction sound?

src/game/Actor.ts
```diff
@@ -144,6 +144,9 @@ export default class Actor {
     readonly index: number;
     readonly props: ActorProps;
     readonly state: ActorState;
+    readonly sound: any;
+    readonly soundStepLeft: any;
```
slwilliams, line 148: Do we really need separate positional audio objects for the different steps? Can't we just get away with one for everything?

slwilliams (commented):
I wasn't able to find anything obvious related to performance issues and on my desktop it wasn't at all noticeable if there was any performance hit.

src/audio/index.ts
xesf: I actually forgot to set the type of the filter, so it was not set as "allpass". Changing the type puts the sound back to the current version.

src/game/Extra.ts
xesf: That's very strange, I can hear them.

src/game/scenery/island/environment/Lightning.ts
xesf: The lightening are a bit lower now than it was before, but there is definitely some issues there.
Sometimes it looks like it does not update the position of the lighting or the camera and the sound is played far away.
At lupin burg near the library is where you can hear pretty well.
I will have a look to see what's happening there.

src/audio/index.ts
xesf: Mainly for actors when they turn, so we can listening to steps or other sounds differently.
Desert Island is a good place to hear it with the cone.
This could potentially be a configuration instead of setting it by default.

src/game/Actor.ts
xesf: Actually not for steps, but for multiple samples. I will try to keep a single one and see if when multiple samples are played for the same actor. We may hear some of the samples being stop before the new one play.

src/cameras/3d.ts
xesf: Documentation advices to attach to the camera. Twinsen sound is not positional, so even if the camera is away, Twinsen sounds will always play the same. Same for the Voices.
I will try to see if adding it to Twinsen its better.

slwilliams (commented):
I still think this is a non trivial regression compared to what we have today and I think this is going to be tricky to get working at a high standard. I think we probably need an option which enables / disables positional audio so we can work on this incrementally. 

High level comments:
- My comment about having two audio sources for the left and right steps has now meant that sound effects get cut off when you start walking. However I think we're taking the wrong approach. We should be in a situation we can play an arbitrary number of samples without needing N audio sources. Not sure the best way to achieve that though. This seems to work fine today without needing 2 audio sources?
- The positional aspect of the audio seems backwards, sounds that are to the left of Twinsen come out of my right speaker.
- The roll off factor and max distance etc need tweaking, I'll post a video showing this.

However if we can get a flag / option in place we can tackle these things in a follow up PR.

xesf:
> I still think this is a non trivial regression compared to what we have today and I think this is going to be tricky to get working at a high standard. I think we probably need an option which enables / disables positional audio so we can work on this incrementally.
> 
> High level comments:
> 
> * My comment about having two audio sources for the left and right steps has now meant that sound effects get cut off when you start walking. However I think we're taking the wrong approach. We should be in a situation we can play an arbitrary number of samples without needing N audio sources. Not sure the best way to achieve that though. This seems to work fine today without needing 2 audio sources?
> * The positional aspect of the audio seems backwards, sounds that are to the left of Twinsen come out of my right speaker.
> * The roll off factor and max distance etc need tweaking, I'll post a video showing this.
> 
> However if we can get a flag / option in place we can tackle these things in a follow up PR.

Thanks for taking the time to properly test it.

* The roll off factor is for now left as the default, only Lightning has a slightly different values so we can hear them a bit better when far away.

* The backwards positional of the audio is fixed and the listener is now attached correctly to Twinsen.

* I have added a flag to control the positional audio and added a param called "audio3d" set to true by default.

The only piece missing now is the multiple Samples. I won't be able to have a single sound with multiple sources as each source is attached to a single buffer, but I will see if I can follow the same approach I did previously, but most likely will be on a separate PR.

slwilliams:
ok plays much better now, couple of things which I think are known issues due to multiple samples not being able to be played at once:
  Twinsen getting hit sounds weird compared to the original (get bob to hit you).
  The jetpack / protopack sounds effects it's playing all the way like it does now seems to loop very quickly.
  Sometimes actors speaking just doesn't work and is delayed by ~15 seconds. Can't repro consistently though so not sure what's going on there...

Thanks for incorporating the flag, I would usually say it should default to false until we achieve parity but it's up to you, don't have a strong opinion here.

slwilliams (approved):

xesf:
> ok plays much better now, couple of things which I think are known issues due to multiple samples not being able to be played at once:
> Twinsen getting hit sounds weird compared to the original (get bob to hit you).
> The jetpack / protopack sounds effects it's playing all the way like it does now seems to loop very quickly.
> Sometimes actors speaking just doesn't work and is delayed by ~15 seconds. Can't repro consistently though so not sure what's going on there...
> 
> Thanks for incorporating the flag, I would usually say it should default to false until we achieve parity but it's up to you, don't have a strong opinion here.

That's done. I have disabled it by default and included an new Options Menu where you can toggle those.

PR #470, opened 2020-11-06
Implement 3D models within inventory

slwilliams:
Still need to add in the object text and there's a reasonable amount of game logic needed for some of the more advanced mechanics e.g. places where slots are reused, items dependent on magic level etc.

**Preview here:** https://pr-470.lba2remake.net

### DIFF
--- a/src/ui/game/Inventory.tsx
+++ b/src/ui/game/Inventory.tsx
@@ -1,14 +1,101 @@
-import React, { useEffect, useState } from 'react';
+import * as THREE from 'three';
+
+import React, { useEffect, useState, useRef } from 'react';
 
 import { SampleType } from '../../game/data/sampleType';
 import { GetInventoryMapping, GetInventoryRows, GetInventoryColumns } from '../../game/data/inventory';
 import '../styles/inventory.scss';
+import {
+    createOverlayScene,
+    createOverlayClock,
+    createOverlayCanvas,
+    createOverlayRenderer,
+    loadSceneInventoryModel,
+} from './overlay';
 
-const inventoryColumns = 7;
-const inventoryRows = 5;
+// fixme: remove global vars
+let clock = null;
+let canvas = null;
+let renderer = null;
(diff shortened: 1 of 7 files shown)


slwilliams:
Test failures are linting for line length. Are we strict on this? looking at the lines that are over the limit they are reasonable and will look much worse if I split them.

agrande:
> Test failures are linting for line length. Are we strict on this? looking at the lines that are over the limit they are reasonable and will look much worse if I split them.

Well, I guess we're not 100% strict, you can always disable this instance with a `// tslint:disable-next-line:max-line-length` if you really think it's more readable like this, but I think the rule is pretty useful overall.

src/ui/game/Inventory.tsx
```diff
@@ -1,14 +1,84 @@
...
+        setModels(models);
+    };
+
+   useEffect(() => {
+       if (clock) {
+            clock.start();
+            for (let i = 0; i < GetInventoryRows(); i += 1) {
+                for (let j = 0; j < GetInventoryColumns(); j += 1) {
+                    const slot = i * GetInventoryColumns() + j;
+                    const inInventory = game.getState().flags.quest[GetInventoryMapping()[slot]] === 1;
```
agrande, line 69: Its one of the lines with a warning, right?
I actually think it can be simplified by for instance taking the quest array out of the loop, like:
`const { quest } = game.getState().flags;`

agrande: The intent of the line limit is not just to bother people with formatting, but also to make sure there's not too much information for our brains to parse in a single line, and the proper fixes are usually more about simplifying the code, rather than formatting it differently.

slwilliams: I go back and forth on this. I lean towards in-lining things these days because it's easier to read without needing state around what variables are what. but tbh in this case it's not too bad so fixed these cases.

agrande (approved):

PR #476, opened 2020-11-09
fix import script folder creation

krampy01:
Running the data import on fresh copy fails as the www/data folder doesn't exist.
Enabling recursive folder creation will ensure that the folders are created correctly.

**Preview here:** https://pr-476.lba2remake.net


slwilliams (approved):
thanks for fixing this!

PR #479, opened 2020-11-12
Implement equipping weapons

slwilliams:
There's some things that need cleaning up here regarding the item IDs, will be sorting that out in a follow up PR.

**Preview here:** https://pr-479.lba2remake.net


slwilliams:
I've hardcoded adding the weapons so it's easier to test, I'll remove that before I merge.

slwilliams:
I also need to take a pass on LBA1 support once I've got LBA2 finished.

src/game/GameState.ts
```diff
@@ -108,6 +109,10 @@ function createQuestFlags() {
     quest[150] = 1;
     quest[152] = 1; // rain
     quest[159] = 256;
+    quest[10] = 1;
+    quest[11] = 1;
+    quest[9] = 1;
+    quest[23] = 1;
```
xesf, line 115: ahh, this reminds me we need to sort this list for LBA1 as well.
Don't worry, I will take this one later.

xesf (approved):

src/game/GameState.ts
slwilliams: Ack thanks

PR #484, opened 2020-11-18
Fixed meca pinguim checks for both LBA1 and LBA2

xesf:
For LBA2 the meca pinguim was always Actor 1, but this is not true for LBA1, where meca pinguim is the very last actor.
This will fix some actors been hidden by mistake in LBA1 (like the Truck Outside the Citadel Island)


slwilliams (approved):

xesf:
Thanks @slwilliams that was fast!

PR #485, opened 2020-11-18
Make padding on inventory relative to screen size

slwilliams:
As the inventory scales the padding scales too, also fixed the alignment with the inventory text box.


xesf:
On Editor mode, the inventory is mostly the normal size and you cannot see the whole inventory area.

xesf (approved):

slwilliams:
hmm ack ok will take a look in a follow up PR because it doesn't look trivial

PR #486, opened 2020-11-19
Fixed isometric Twinsen position between scenes

xesf:
Issue was more evident on LBA1 where Twinsen was not placed correctly on the new scene.


slwilliams:
Haven't actually tested since the preview instance doesn't seem to be working but since this is the unusual case when we can't find a zone to snap to it's probably ok.

slwilliams (approved):

PR #488, opened 2020-12-06
Grey Buildings set: conveyor belt asset set

Peterdevos:
Made 3d replacements for the conveyor belt assets used in the warehouse grids

**Preview here:** https://pr-488.lba2remake.net


slwilliams (approved):
looks great!

PR #490, opened 2020-12-24
Initial magic ball implementation

slwilliams:
Initial implementation of the magic ball, notably:

- Collision and bouncing works for both iso and 3d envs.
- Implemented hitting actors (and fixed a bug with sprite actors).
- Implemented logic for magic (number of bounces etc.) and level progressions.
- Refactored the inventory items stuff fixed some TODOs for me.

Still TODO:

- The ball is probably a bit too large in the 3d env (looks good in iso though) and bounces for too long, some tweaking is needed here.
- Fire level sound effects (they're different).
- Picking up keys
- VR implementation.

Code notes:

For some reason I've got some crazy offsets to convert the triangle coordinates into world coordinates (+80 and + 40). I have no idea where they're coming from and found them by trial and error. Any insight here would be appreciated:

```
const triangleToWorld = (p: THREE.Vector3) => {
    return new THREE.Vector3(
       WORLD_SIZE_M2 * (section.x + 1) - ((p.x - 1) / GRID_SCALE) + 80,
       p.y,
       (p.z / GRID_SCALE) + section.z * WORLD_SIZE_M2 + 40,
    );
};
```

**Preview here:** https://pr-490.lba2remake.net

### DIFF
--- a/src/game/Scene.ts
+++ b/src/game/Scene.ts
@@ -6,6 +6,7 @@ import Actor, { ActorProps, ActorDirMode } from './Actor';
 import Point, { PointProps } from './Point';
 import Zone, { ZoneProps } from './Zone';
 import Extra from './Extra';
+import MagicBall from './MagicBall';
 import { loadScripts } from '../scripting';
 import { killActor } from './scripting';
 import { createFPSCounter } from '../ui/vr/vrFPS';
@@ -64,6 +65,7 @@ export default class Scene {
     readonly scenery: Scenery;
     sideScenes: Map<number, Scene>;
     extras: Extra[];
+    magicBall: MagicBall;
     isActive: boolean;
     isSideScene: boolean;
     zoneState: {
@@ -282,6 +284,9 @@ export default class Scene {
         for (const extra of this.extras) {
             extra.update(this.game, this, time);
         }
+        if (this.magicBall) {
+            this.magicBall.update(time);
+        }
         if (params.editor && this.isActive) {
             for (const point of this.points) {
                 point.update(this.camera);
@@ -414,6 +419,19 @@ export default class Scene {
     addExtra(extra: Extra) {
         this.extras.push(extra);
         this.addMesh(extra.threeObject);
+    }
+
+    addMagicBall(magicBall: MagicBall) {
+        if (this.magicBall) {
+            this.removeMagicBall();
+        }
+        this.magicBall = magicBall;
+        this.addMesh(magicBall.threeObject);
+    }
+
+    removeMagicBall() {
+        this.removeMesh(this.magicBall.threeObject);
+        this.magicBall = null;
     }
 
     removeExtra(extra: Extra) {
--- a/src/game/data/sampleType.ts
+++ b/src/game/data/sampleType.ts
@@ -1,6 +1,9 @@
 
 export const SampleType = {
+    MAGIC_BALL_THROW: 0,
     OBJECT_FOUND: 6,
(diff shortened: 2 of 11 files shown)


src/game/Scene.ts
```diff
@@ -64,6 +65,7 @@ export default class Scene {
     readonly scenery: Scenery;
     sideScenes: Map<number, Scene>;
     extras: Extra[];
+    magicBall: MagicBall;
```
xesf, line 68: Just a remark, this was probably done by design, but MagicBall is an Extra type and we are separating it.
I don't see any problem on that, specially that MagicBall has the bounce behaviour all the other Extras don't have, but all the projectile and hit routines will be the same expect the force, speed and angles.

src/game/loop/hero.ts
```diff
@@ -389,23 +389,22 @@ function processActorMovement(
         }
         if (controlsState.weapon === 1) {
             switch (game.getState().hero.equippedItemId) {
-                // TODO(scottwilliams): Extract inventory item IDs out into a const list.
-                case 1:
+                case LBA2Items.MAGIC_BALL:
```
xesf, line 392: Magicball is not automatically selected when you found the Magicball. I believe in the original you did not had to select them in the inventory, it will automatically select the magicball.

src/game/scenery/isometric/IsoSceneryPhysics.ts
```diff
@@ -24,6 +24,61 @@ export default class IsoSceneryPhysics {
...
+                            return new THREE.Vector3(1 * Math.sign(dir.x), 0, 0).normalize();
+                        }
+                        return new THREE.Vector3(0, 0, 1 * Math.sign(dir.z)).normalize();
+                    }
+                }
+            }
+        }
+        return null;
+    }
+
```
xesf, line 81: The magicball seems slow with default values compared with the original. Not sure values need to be tune per Type (iso or island).

The ball also seems big on Island, not only on Iso.

src/game/data/sampleType.ts
```diff
@@ -1,6 +1,9 @@
 
 export const SampleType = {
+    MAGIC_BALL_THROW: 0,
     OBJECT_FOUND: 6,
+    MAGIC_BALL_BOUNCE: 7,
+    MAGIC_BALL_STOP: 8,
```
xesf, line 6: I couldn't test on LBA1, not sure these are the same.
Since we have some issues with the scripting I need to trigger some flags to allow me to test this in LBA1.

xesf (approved):

src/game/Scene.ts
slwilliams: ack yea I did think about this, but decided that the amount of logic these share is very little and that it would over complicate the extras class for not much value.

src/game/loop/hero.ts
slwilliams: ack yea this is on my todo list

src/game/scenery/isometric/IsoSceneryPhysics.ts
slwilliams: yea also on my todo list, need to tweak a few of the params but will need a bit of experimentation

src/game/data/sampleType.ts
slwilliams: ack yea I didn't test in LBA 1 yet either, my approach at the moment is to get things working in lba 2 and then port to lba 1 after.

PR #495, opened 2021-01-03
Fix iso collision a bit and implement fireball sound effect

slwilliams:
Fixes a few random things:

- Ralph now runs correctly out of the Trulu cave, the DIFF of 1/128 was making his bounding box just slightly too high and clipping the top of the door. Slightly hacky fix but seems to work ok.
- Implement using the correct sound effect for the fire magic ball.
- Slightly tighten up the spike collision detection to match original implementation.

**Preview here:** https://pr-495.lba2remake.net

### DIFF
--- a/src/game/MagicBall.ts
+++ b/src/game/MagicBall.ts
@@ -126,7 +126,11 @@ export default class MagicBall {
     }
 
     throw(angle, behaviour) {
-        this.scene.actors[0].playSample(SampleType.MAGIC_BALL_THROW);
+        if (this.game.getState().hero.magicball.level < 4) {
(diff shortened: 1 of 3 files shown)


slwilliams:
Nah ok this has broken a bunch of other things with iso collision, will try and resolve.

slwilliams:
ok found a much nicer way of doing this, rather than moving the entire actor box up by 1/128 I just move the bottom of the actor box. This means the top of Ralph's bounding box no longer clips the top of the door, but all the other corner cases (like running up steps) still work as expected.

xesf (approved):

PR #499, opened 2021-01-31
New layout replacements

Peterdevos:
Mostly DI. Should have all essential layout replacements covered now. Final GB (books) layout made and added as well.

**Preview here:** https://pr-499.lba2remake.net


slwilliams (approved):
had a play around, looks great!

PR #505, opened 2021-02-21
Fix HIT script implementation.

slwilliams:
This fixes the button in the museum not working and was likely broken when we did that migration to add Actor state in the state object.

**Preview here:** https://pr-505.lba2remake.net


agrande (approved):
Weird that I didn't find this one when refactoring.

xesf:
That is like finding a needle into a haystack 🥇

PR #510, opened 2021-03-29
Implement point offset ability, fix Twinsen exiting Trulu cave.

slwilliams:
**Preview here:** https://pr-510.lba2remake.net

### DIFF
--- a/src/game/Scene.ts
+++ b/src/game/Scene.ts
@@ -25,6 +25,7 @@ import Island from './scenery/island/Island';
 import { Time } from '../datatypes';
 import { processPhysicsFrame } from './loop/physics';
 import { SpriteType } from './data/spriteType';
+import { LBA2PointOffsets } from './scripting/data/lba2/points';
 
 export interface SceneProps {
     index: number;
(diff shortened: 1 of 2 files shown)


agrande (approved):
Hmm, I'm surprised we should need this at all.
I'd like if we could fix the root issue somehow...
Which script command is causing the issue? Maybe our distance conversions are not properly calibrated?
But since this shouldn't affect anything else, and makes the game more playable, I think we should proceed with it.

xesf (commented):
There was some threshold for distance to be "accepted" and I think it was exactly 0.5, but it could be enough to cause the Bounding Boxes to intercept with Zones causing the hang.

xesf:
Could be these two functions

export function getDistance(value) {
    return (value * distanceThreeJs) / 500;
}

export function getDistanceLba(value) {
    return (value * 500) / distanceThreeJs;
}

https://github.com/agrande/lba2remake/blob/1555b42fdd87173440935db8d4b94619314e478d/src/utils/lba.ts#L89

PR #511, opened 2021-03-30
Fix bodyIndex and entityIndex changing outside of the actor loop.

slwilliams:
This fixes Twinsen losing his Tunic when he teleports to a new area without having picked up the magic ball as well as transitioning to Twinsen+Zoe upon exiting the Trulu cave.

**Preview here:** https://pr-511.lba2remake.net

### DIFF
--- a/src/game/loop/hero.ts
+++ b/src/game/loop/hero.ts
@@ -61,7 +61,7 @@ function handleBodyChanges(game: Game, scene: Scene, hero: Actor) {
     const equippedItem = game.getState().hero.equippedItemId;
     if (equippedItem < 0) {
         // Corner case for when Twinsen hasn't yet picked up the magic ball.
-        if (game.getState().flags.quest[4]) {
+        if (game.getState().flags.quest[LBA2Items.TUNIC]) {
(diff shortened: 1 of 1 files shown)


src/game/loop/hero.ts
```diff
@@ -51,8 +53,19 @@ export function updateHero(game: Game, scene: Scene, hero: Actor, time: Time) {
...
 function handleBodyChanges(game: Game, scene: Scene, hero: Actor) {
+    if (hero.props.dirMode !== ActorDirMode.MANUAL) {
+        hero.setBody(scene, hero.props.bodyIndex);
+        return;
+    }
+
     const equippedItem = game.getState().hero.equippedItemId;
     if (equippedItem < 0) {
+        // Corner case for when Twinsen hasn't yet picked up the magic ball.
+        if (game.getState().flags.quest[4]) {
```
xesf, line 64: We should use a descriptive GameFlagType type here.

xesf (approved):

src/game/loop/hero.ts
slwilliams: good idea, fixed.

PR #520, opened 2021-04-12
Camera Zones

agrande:
Only enabled in cinematic mode, or when Twinsen dirMode is not manual, for now.
Will try to apply it in regular gameplay in a nice way in a future PR.

**Preview here:** https://pr-520.lba2remake.net

### DIFF
--- a/src/cameras/3d.ts
+++ b/src/cameras/3d.ts
@@ -97,15 +97,21 @@ function processFollow3DMovement(
     cameraPos.applyMatrix4(hero.threeObject.matrixWorld);
     scene.scenery.physics.processCameraCollisions(cameraPos);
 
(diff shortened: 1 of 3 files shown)


slwilliams:
Playing the Trulu exit scene through the camera still does some weird stuff I guess this is because we're tracking Twinsen rather than leaving the camera static like in the original game. This results in some strange camera angles that don't look great (clipping through things etc.).

Not sure what's best to do but this is a decent start.

The other thing we need is the CAMERA_FOLLOW life script command implemented e.g. so the camera follows the boat when it leaves the harbour etc.

slwilliams (approved):

agrande:
Yeah, definitely not perfect, I'll iterate on this.

PR #521, opened 2021-04-13
Small adjustments

xesf:
* Improved Bonus Trajectory
* Improved Bonus not go beyond scene
* Reduce Magic Ball Size

**Preview here:** https://pr-521.lba2remake.net

### DIFF
--- a/src/game/MagicBall.ts
+++ b/src/game/MagicBall.ts
@@ -51,7 +51,7 @@ export default class MagicBall {
             true, /* isBillboard */
             this.scene.is3DCam,
         );
-        sprite.threeObject.scale.multiplyScalar(this.scene.vr ? 1.5 : 1);
+        sprite.threeObject.scale.multiplyScalar(1);
         this.threeObject.add(sprite.threeObject);
         this.threeObject.name = 'magicball';
         this.threeObject.visible = true;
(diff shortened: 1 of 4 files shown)


src/game/MagicBall.ts
```diff
@@ -51,7 +51,7 @@ export default class MagicBall {
             true, /* isBillboard */
             this.scene.is3DCam,
         );
-        sprite.threeObject.scale.multiplyScalar(1.5);
+        sprite.threeObject.scale.multiplyScalar(this.scene.vr ? 1.5 : 1);
```
agrande, line 54: Not sure I understand this change. What does VR have to do with it?
As far as I understand, the magicball size issue is related to canvas size, it will appear differently on different resolutions. No magic number will change that.

xesf: VR was the cause of the initial scale up of the sprites, reason I left that.
Sprites are not currently been resized based on canvas size, perhaps we should as in some small devices like playing on the phone the sprites will look bigger without scaling down.

slwilliams (approved):

src/game/MagicBall.ts
xesf: I removed the scale completely for now until a proper solution is made for all sprites.

PR #525, opened 2021-04-18
Implement CAM_FOLLOW life command

slwilliams:
Rather basic implementation. At the moment the boat scene to Desert island is definitely better but it's still not close to what the original game does. I'm not sure if we want to aim for that, if we do the cam zone stuff needs work.

I've included the flag changes to make it easy to test (just teleport to the harbour and you'll already have a ticket and it'll be sunny) I'll revert these before merging.

**Preview here:** https://pr-525.lba2remake.net


xesf:
Looks awesome!!! I really like it.

xesf (approved):

PR #527, opened 2021-04-18
Animation action throw with actor hit

xesf:
**Preview here:** https://pr-527.lba2remake.net

### DIFF
--- a/src/game/Extra.ts
+++ b/src/game/Extra.ts
@@ -306,6 +306,12 @@ export default class Extra {
 
         this.physics.position.add(trajectory);
         this.threeObject.position.copy(this.physics.position);
+        this.threeObject.quaternion.copy(
+            new THREE.Quaternion().setFromUnitVectors(
+                new THREE.Vector3(0, 0, 1),
+                trajectory.normalize(),
+            ),
+        );
     }
 
     update(game: Game, scene: Scene, time: Time) {
--- a/src/game/Extra.ts
+++ b/src/game/Extra.ts
@@ -10,6 +10,7 @@ import Scene from './Scene';
 import Game from './Game';
 import { Time } from '../datatypes';
 import { getParams } from '../params';
+import { loadInventoryModel } from '../model/inventory';
 // import { createBoundingBox } from '../utils/rendering';
 
 const ExtraFlag = {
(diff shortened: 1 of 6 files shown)


src/game/loop/zones.ts
```diff
@@ -442,7 +442,7 @@ function BONUS(game: Game, scene: Scene, zone: Zone, hero: Actor, time: Time) {
                 game,
                 scene,
                 position,
-                destAngle,
+                THREE.MathUtils.degToRad(78),
```
slwilliams, line 445: This seems suspicious. Why is this hardcoded to 78?

slwilliams (commented):
Searching in chests etc. doesn't seem to work very well anymore.

slwilliams (commented):
ok after realising what this actually does and testing it properly. Couple of points:

- I think there's a timing delay / offset here since the projectiles don't fire at the same time as the sound effects.
- I think there's also a model offset we're missing because the projectiles don't seem to come from the e.g. gun that's being shot but slightly to the left / backwards.

but this is very cool regardless!

src/game/loop/zones.ts
xesf: That part is actually a value passed in the Bonus zones as an argument. Since the angle for other projectiles comes from the anim actions, but for bonus is manually triggered per zone.

I still need to check if the value it self is correct though

xesf:
Pushed a change for it.
Seems a bit better, although not perfect.

There is a 0.15 off on X coordinate for some reason, but I left it without any magic numbers for now.

src/game/Extra.ts
```diff
@@ -83,22 +85,71 @@ export default class Extra {
     spawnTime: number;
     speed: number;
     weight: number;
+    throwAngle: number;
+    hitStrength: number;
     hasCollidedWithActor: boolean;
 
     static async load(
```
agrande, line 92: Is the load method still used anywhere?

agrande (approved):

slwilliams (approved):

src/game/Extra.ts
```diff
@@ -219,6 +296,24 @@ export default class Extra {
...
+        trajectory.applyEuler(new THREE.Euler(0, this.physics.temp.angle, 0, 'XZY'));
+
+        this.physics.position.add(trajectory);
+        this.threeObject.position.copy(this.physics.position);
+        this.threeObject.quaternion.copy(
+            new THREE.Quaternion().setFromUnitVectors(
+                new THREE.Vector3(0, 0, 1),
+                trajectory.normalize(),
+            ),
+        );
```
agrande, line 314: Try to avoid allocations in functions that are called at every frame...
These computations can probably be made without creating new Vector3s / Quaternions / Eulers.
One common pattern I use is have a global variable in UPPERCASE in the file that I use for temporary calculations.

agrande: We pay a double performance price for allocations:
- the allocation itself takes some time
- then if there's enough of them over the course of several frames, it will trigger the garbage collector to run and cause dropped frames

src/game/Extra.ts
xesf: I remove it, was no longer need it.

src/game/Extra.ts
xesf: Yeah, I was thinking the same when I was already in bed 😅.
Will do that fiz before merging.

PR #534, opened 2021-04-21
Cancel jump in mid-air with down arrow key

agrande:
**Preview here:** https://pr-534.lba2remake.net


slwilliams (approved):
huh never knew this was a thing.

PR #536, opened 2021-04-21
More 3D layout replacements

SendellX:
All these layouts are used in the spacecraft area in the Temple of Bu. I'm almost done replacing the layouts for the whole Temple.

**Preview here:** https://pr-536.lba2remake.net


xesf (approved):
Looking great. Nice work!!.

Btw, you forgot to add the source assets.

PR #537, opened 2021-04-22
Carrying actors

agrande:
**Preview here:** https://pr-537.lba2remake.net


src/game/loop/hero.ts
```diff
@@ -257,7 +257,7 @@ const MEDIUM_FALL_HEIGHT = 2;
 const SMALL_FALL_HEIGHT = 0.3;
 
 function processFall(scene: Scene, hero: Actor) {
-    let distFromFloor = hero.state.distFromGround;
+    let distFromFloor = hero.state.distFromFloor;
```
agrande, line 260: @slwilliams I had to change this to make it work. It doesn't seem to have any negative consequences, and I assume it was a typo, but tell me if I broke something.

xesf:
There seem to be some weird behaviour when Twinsen is carried by a platform.
If you use Up/Down keys, both objects moves at a higher rate together with Twinsen. Left of Right don't do anything as expected.
This is the same issue you have posted a video on Slack so I am not sure this is a known issue to be looked after.

agrande:
> There seem to be some weird behaviour when Twinsen is carried by a platform.
> If you use Up/Down keys, both objects moves at a higher rate together with Twinsen. Left of Right don't do anything as expected.

Does that happen on the cart in Temple of Bu? I didn't notice it. I'm not entirely sure what the problem looks like.
How can I reproduce it exactly?

> This is the same issue you have posted a video on Slack so I am not sure this is a known issue to be looked after.

Yes, it is a known issue, my intent at first was just to make the cart work in Temple of Bu, the other stuff were an afterthought, and I didn't thoroughly investigate yet what prevents them from working. Not sure if I'll do this here or in a future PR.

src/game/loop/hero.ts
slwilliams: nah that looks good to me

src/game/loop/physics.ts
```diff
@@ -41,14 +46,71 @@ function processActorPhysics(game: Game, scene: Scene, actor: Actor, time: Time)
...
+        }
+    }
+}
+
+const TMP_POS = new THREE.Vector3();
+const TMP_Q = new THREE.Quaternion();
+const TMP_EULER = new THREE.Euler();
+
+function processCarriedPosition(scene: Scene, actor: Actor) {
+    if (actor.state.isCarriedBy !== -1) {
```
slwilliams, line 64: an early return here is much nicer to read:

if (actor.state.isCarriedBy === -1) {
  actor.state.isCarried = false;
  return;
}

rest of the code....

slwilliams (approved):

xesf:
Looking good now. Have tried on different platforms and its working as expected.

Just notice one tiny thing that could be fixed at a later stage.
In the Sewers platform, since it has a rounded shape, it seems the Y position does not follow the Platform model Y position, so Twinsen feets are inside it.

xesf:
There is another thing, I believe known issue where you get stuck with no control on Twinsen, if you try to enter the Sewers using the platform. Same happen for sprite platforms inside the sewers. Maybe it uses a different script command not implemented yet.

agrande:
> In the Sewers platform, since it has a rounded shape, it seems the Y position does not follow the Platform model Y position, so Twinsen feets are inside it.

> There is another thing, I believe known issue where you get stuck with no control on Twinsen, if you try to enter the Sewers using the platform. Same happen for sprite platforms inside the sewers. Maybe it uses a different script command not implemented yet.

Yeah, I noticed those 2 issues, not sure how to deal with them, but I'll merge this as it is for now and see how I can fix them in future PRs.

PR #541, opened 2021-04-24
Shading improvements

agrande:
Made some fixes on shading using the properly parsed polyType property:

- Solid shading is when we just need to use a plain palette color without it being affect by lighting conditions. Examples: character eyes, most of the spheres (Twinsen's hair, etc).
- Flat shading is when the normals being used are computed face normals instead of vertices normals (this means models look more edgy instead of the gouraud shading). Examples: cart and arrows in Temple of Bu.
- Textures test was wrong, any polyType > 7 has texture. Probably still need some improvements for some specific textured shader cases that we're not handling properly yet.

The polyType property along with the PolygonType enum will allow us to implement the other original shaders more easily (in a future PR).

**Preview here:** https://pr-541.lba2remake.net

### DIFF
--- a/src/model/geometries.ts
+++ b/src/model/geometries.ts
@@ -8,7 +8,7 @@ import FRAG_TEXTURED from './shaders/textured.frag.glsl';
 
 import {loadPaletteTexture, loadSubTextureRGBA} from '../texture';
 import {compile} from '../utils/shaders';
-import { WORLD_SIZE } from '../utils/lba';
+import { WORLD_SIZE, PolygonType } from '../utils/lba';
 import Lightning from '../game/scenery/island/environment/Lightning';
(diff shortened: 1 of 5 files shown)


xesf (approved):
Nice one!!

It also brought some changes to the LBA1 models.
Although the colours are off, the overall result seems better.

I've added
bufferGeometry.computeVertexNormals()
for LBA1 and I think it can be removed for now after this PR.

agrande:
> I've added
> 
> bufferGeometry.computeVertexNormals()
> 
> for LBA1 and I think it can be removed on this PR.

I don't know if it can be skipped as vertex normals are not the same as face normals. If we need to compute normals for the Gouraud shading in LBA1, we'll need a different algorithm than this one.

PR #544, opened 2021-04-24
Fix dirMode FOLLOW implementation to correctly pass target actor

slwilliams:
This means enemies now actually rotate to face Twinsen and shoot at him properly.

**Preview here:** https://pr-544.lba2remake.net


agrande (approved):
Lgtm

PR #546, opened 2021-04-25
Add missing Poly Type to LBA1 models

xesf:
**Preview here:** https://pr-546.lba2remake.net

### DIFF
--- a/src/model/geometries.ts
+++ b/src/model/geometries.ts
@@ -261,6 +261,10 @@ export function loadMesh(
                 new THREE.BufferAttribute(new Uint8Array(lineBones), 1)
             );
 
+            if (isLBA1) {
+                linebufferGeometry.computeVertexNormals();
+            }
+
             const lineSegments = new THREE.LineSegments(linebufferGeometry, material);
             lineSegments.onBeforeRender = Lightning.applyUniforms;
             lineSegments.name = 'lines';
--- a/src/utils/lba.ts
+++ b/src/utils/lba.ts
@@ -26,7 +26,9 @@ const getAngleValues = () => {
     return { midValue, maxValue };
 };
 
-export const PolygonType = {
+const isLBA1 = getParams().game === 'lba1';
+
+export const PolygonType = !isLBA1 ? {
     SOLID: 0,
     FLAT: 1,
     TRANS: 2,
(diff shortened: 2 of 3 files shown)


slwilliams (approved):

src/utils/lba.ts
```diff
@@ -53,6 +55,18 @@ export const PolygonType = {
...
+    BOPPER: 2,
+    MARBLE: 3,
+    TELE: 4,
+    TRANS: 5,
+    TRAME: 6,
+    GOURAUD: 7,
+    DITHER: 8,
+    GOURAUD_TABLE: 9,
+    DITHER_TABLE: 10,
 };
```
agrande, line 70: Wouldn't it be better to make a different enum for LBA1, and use a function to perform a mapping in the parser, so that the game engine only sees LBA2 polygon types?
That's assuming we can map everything from LBA1 polyTypes to LBA2 polyTypes, but if we can, maybe we should have 3 enums: `PolyType`, `PolyType1` and `PolyType2`.
The point is that shader code is gonna be using the polyType attribute, and we don't want to have `if (LBA1)` in glsl shaders if we can avoid it.

src/model/geometries.ts
```diff
@@ -261,6 +261,10 @@ export function loadMesh(
                 new THREE.BufferAttribute(new Uint8Array(lineBones), 1)
             );
 
+            if (isLBA1) {
+                linebufferGeometry.computeVertexNormals();
```
agrande, line 265: From the look of it, it doesn't seem like the normals are computed the way we need.
I think for flat shading we want to compute face normals, which is what this seems to do, and for gouraud shading we want to compute the face normals of all faces that connect at a specific vertex, and then average those face normals into the final vertex normal.

agrande: Assuming LBA1 doesn't provide the normals in the model format.

src/utils/lba.ts
xesf: Yep, it's a great idea. Totally forgot we could do it during parsing and it makes everything else less game dependent.
I will create another PR with this change.

Not every type exist in LBA2, but we can add additional ones with different opcodes.

src/model/geometries.ts
xesf: I was reading a bit about it yesterday and it indeed seems we need face normals.
And yes, LBA1 does not have the normals in the model format.

Using computeVertexNormals is more of a workaround for now, so that the game looks a bit similar to original even though very low poly looking. I expect this could be removed if we manage to get the face normals to work.

PR #557, opened 2021-04-30
Fix shader compilation issue on webgl1

agrande:
**Preview here:** https://pr-557.lba2remake.net


xesf (approved):
Working fine now for both desktop and mobile.

PR #558, opened 2021-05-01
WAIT DOOR script command

xesf:
**Preview here:** https://pr-558.lba2remake.net

### DIFF
--- a/src/game/Actor.ts
+++ b/src/game/Actor.ts
@@ -73,6 +73,7 @@ interface ActorPhysics {
         destination: THREE.Vector3;
         angle: number;
         destAngle: number;
+        doorPosition?: [number, number, number];
     };
     carried: {
         position: THREE.Vector3;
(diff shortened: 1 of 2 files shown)


slwilliams (approved):

agrande:
What is this command used for?

xesf:
> What is this command used for?

Wait door is actually what implements the opening of the doors.
when OPEN_LEFT/RIGHT/UP/DOWN is used, the script continues and checks if the door was actually open in the WAIT_DOOR. Something like the WAIT_ANIM where it waits for animation to finish, this one waits for the door to finish open.

This was causing some weird behaviour in the scripts, for example the LBA1 Outside Prison fences which has more conditions to open the fences.

agrande (approved):

agrande:
Door in the temple of Bù (the ones you open when on the train) broke.
Only one of them opens.

xesf:
> Door in the temple of Bù (the ones you open when on the train) broke.
> Only one of them opens.

Yeah, saw it now on LBA1 as well.
Looks like not all doors implement the WAIT_DOOR command.
I have a quick fix for it.

PR #559, opened 2021-05-01
WORKAROUND: Swap XZ axis during collision test if actor has rotated b…

xesf:
…etween threshold

**Preview here:** https://pr-559.lba2remake.net

### DIFF
--- a/src/game/loop/physics.ts
+++ b/src/game/loop/physics.ts
@@ -162,8 +162,7 @@ function processCollisionsWithActors(scene: Scene, actor: Actor) {
     // Swap X and Z axis if actor has rotation
     // TODO to improve how we handle AABB rotation
     const angleDeg = THREE.MathUtils.radToDeg(actor.physics.temp.angle);
-    if ((angleDeg < 45 && angleDeg > 315) ||
-        (135 < angleDeg && angleDeg <= 225)) {
+    if (!(angleDeg < 45 && angleDeg >= 315)) {
         const x1 = ACTOR_BOX.min.x;
         const x2 = ACTOR_BOX.max.x;
         ACTOR_BOX.min.setX(ACTOR_BOX.min.z);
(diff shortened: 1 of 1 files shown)


agrande:
Doesn't seem to work for me (tested on the garbage truck in LBA1 and Temple of Bù train in LBA2).
Also, looking at the code, I think it's probably gonna be swapping back-and forth all the time.
What we probably need to do is to measure the long side with the initial orientation and then on every frame decide where to put the long side depending on how the current orientation compares with the original.
Also, I would get the angle from the quaternion instead of using this angle value, the angle here can sometimes get out of sync for some reason, and the one in the quaternion is more likely to be normalised to the proper range.

agrande (changes requested):
Doesn't seem to work for me (tested on the garbage truck in LBA1 and Temple of Bù train in LBA2).

Also, looking at the code, I think it's probably gonna be swapping back-and forth all the time.
What we probably need to do is to measure the long side with the initial orientation and then on every frame decide where to put the long side depending on how the current orientation compares with the original.

Also, I would get the angle from the quaternion instead of using this angle value, the angle here can sometimes get out of sync for some reason, and the one in the quaternion is more likely to be normalised to the proper range.

xesf:
Strange it did not work for you. Did the Truck in LBA1 get stuck on the fences?
Or you are guiding it based on the BB Mesh? The mesh orientation is actually not changed, since this patch is only done during actor collision.

I will try to change the code based on the other suggestions.

agrande:
> Strange it did not work for you. Did the Truck in LBA1 get stuck on the fences?
> Or you are guiding it based on the BB Mesh? The mesh orientation is actually not changed, since this patch is only done during actor collision.

Actually, about this, no, it didn't get stuck on the fences.
I was looking at the BB mesh. I didn't realise the patch is only done during actor collision.

Reading the code again, I think it's fine.

agrande (approved):

agrande:
Careful about the LGTM warning though, there's probably something there.

PR #561, opened 2021-05-01
Don't play additional samples when already playing one

slwilliams:
This changes to keep track of samples on a per actor basis i.e. we allow the same sample to be concurrently played but only once per actor. This allows two enemies to shoot at the same time with the same sample and we'll hear both, but in the case of the jetpack we don't play that sample twice since it's the same actor (Twinsen).

This doesn't change the 3D positional audio stuff, need to look at that some more. We also still aren't looping samples correctly (there's still a slight gap at the end of the jetpack sample playback, still trying to work out how best to resolve that.

I've also fixed the case where we end up with the theme music playing twice at the same time which has been annoying me for a while.

**Preview here:** https://pr-561.lba2remake.net

### DIFF
--- a/src/audio/index.ts
+++ b/src/audio/index.ts
@@ -15,6 +15,7 @@ declare global {
 
 const THEME_MENU = 'THEME_MENU';
 const samples = {};
+const samplesPerActor = {};
 
 function createAudioContext() {
     window.AudioContext = window.AudioContext || window.webkitAudioContext; // needed for Safari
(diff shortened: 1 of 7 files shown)


src/audio/index.ts
```diff
@@ -204,11 +205,16 @@ export function createAudioManager(state) {
...
+            actorIndex: number = -1,
             volume: number = state.config.soundFxVolume
         ) => {
             const sampleSource = createSampleSource(context, volume);
             sampleSource.play(index, frequency, loopCount);
             samples[index] = sampleSource;
+            if (!samplesPerActor[actorIndex]) {
+                samplesPerActor[actorIndex] = {};
+            }
+            samplesPerActor[actorIndex][index] = sampleSource;
```
xesf, line 217: Looking at this it just made me realise the issue Adrian notice in VR may be caused by this.

Both samples and samplesPerActor are never released. The more we play, the more samples both objects will be referencing. The decoded sample is cached in 

I will suggest to create a release function that will delete the contents of these objects and called during scene change in the Scene Manager release function.

xesf (changes requested):

xesf:
> We also still aren't looping samples correctly (there's still a slight gap at the end of the jetpack sample playback, still trying to work out how best to resolve that.
Those are triggered per animation frame.
Perhaps we are not stopping the previous sample before starting the new one.

agrande (approved):

xesf (approved):
I've approved so we can merge it.
I will do the sample release based on my comments.

PR #566, opened 2021-05-02
More LBA1 Related Fixes

xesf:
* Brick samples not correct when on Isometric Island mode (islands grouped together)
* Update sample types for LBA1 which where playing the wrong samples
* Magicball sprite update for LBA1

**Preview here:** https://pr-566.lba2remake.net

### DIFF
--- a/src/game/data/sampleType.ts
+++ b/src/game/data/sampleType.ts
@@ -28,6 +28,4 @@ const LBA2SampleType = {
     TWINSEN_LANDING: 129,
 };
 
-export const getSampleType = () => {
-    return isLBA1 ? LBA1SampleType : LBA2SampleType;
-};
+export default isLBA1 ? LBA1SampleType : LBA2SampleType;
--- a/src/game/data/sampleType.ts
+++ b/src/game/data/sampleType.ts
@@ -1,5 +1,21 @@
+import { getParams } from '../../params';
 
-export const SampleType = {
+const isLBA1 = getParams().game === 'lba1';
+
+const LBA1SampleType = {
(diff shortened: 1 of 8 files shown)


slwilliams (approved):

src/game/loop/animAction.ts
```diff
@@ -168,7 +168,7 @@ export const ZV = unimplemented();
 export const LEFT_STEP = (_action: AnimAction, { actor, scene, animState }: AnimActionContext) => {
     const floorSound = animState.floorSound;
     if (floorSound !== undefined && floorSound !== -1) {
-        const offset = scene.props.isIsland ? 30 : isLBA1 ? 126 : 60;
+        const offset = isLBA1 ? 126 : scene.props.isIsland ? 30 : 60;
```
agrande, line 171: I would add some parentheses here. I always forget the precedence rules for nested ternary expressions so I'm not entirely sure how to read this.

src/game/data/sampleType.ts
```diff
@@ -11,3 +27,7 @@ export const SampleType = {
     ACTOR_DYING: 14,
     TWINSEN_LANDING: 129,
 };
+
+export const getSampleType = () => {
```
agrande, line 31: Do we need to export a function? Can't we just export a variable?

agrande (approved):

src/game/data/sampleType.ts
xesf: Was to keep it consistent like the inventory functions but I can export a default variable instead.

PR #567, opened 2021-05-02
Release samples awhile reloading a new scene

xesf:
Still need to test it with VR.

@slwilliams that's the fix I mention on your previous PR.

**Preview here:** https://pr-567.lba2remake.net

### DIFF
--- a/src/audio/index.ts
+++ b/src/audio/index.ts
@@ -138,10 +138,12 @@ export function createAudioManager(state) {
         },
         releaseSamples: () => {
             Object.keys(samples).forEach((key) => {
+                samples[key].stop();
                 delete samples[key];
             });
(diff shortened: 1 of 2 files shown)


agrande:
Ok, I'll test this now.

agrande:
> Ok, I'll test this now.

It didn't come as fast, but I encountered some audio issues after playing for a while. There was some stuttering at some point in a scene with many samples (not as much as before, and it got better after switching scenes, but still).
Also, all in-game audio stopped playing after some time, and I had to reload the game to get it back.

xesf:
Needs a deeper look than. I will try tomorrow to have a look with the VR and see if I can find what's happening.
But good that there are some improvements.

agrande (approved):
This is an improvement already, we can always improve it more in a future PR.

PR #571, opened 2021-05-03
Fixup music playing to behave like the original game

slwilliams:
The mechanics of how the original game plays music are more complex than I realised:

- If you enter a new scene and no music is playing, play the music for that scene.
- If you enter a new scene and music is already playing, let it continue to play until finish, but queue up the current scene music and play that once the currently playing piece has finished.
- If you walk through multiple scenes only queue up the most recent scene music.
- If you walk out of a scene and back in again, don't queue up the same piece of music twice, or play the other scene piece.

Therefore I've added the ability to "queue" up music tracks. Think I've covered all the cases.

**Preview here:** https://pr-571.lba2remake.net

### DIFF
--- a/src/audio/index.ts
+++ b/src/audio/index.ts
@@ -41,6 +41,8 @@ export function createAudioManager(state) {
         listener.rotateY(Math.PI);
     }
 
+    let queuedMusic = -1;
+
     return {
         context,
         menuContext,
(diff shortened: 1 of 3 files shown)


agrande (approved):
Nice findings!

PR #572, opened 2021-05-03
Enable VR for LBA1

xesf:
**Preview here:** https://pr-572.lba2remake.net

### DIFF
--- a/src/controls/vr/mappings.ts
+++ b/src/controls/vr/mappings.ts
@@ -2,6 +2,7 @@ import { MotionController } from '@webxr-input-profiles/motion-controllers';
 import { BehaviourMode } from '../../game/loop/hero';
 import Game from '../../game/Game';
 import { getParams } from '../../params';
+import { LBA1Items, LBA2Items } from '../../game/data/inventory';
 
 interface BtnMapping {
     btn: string;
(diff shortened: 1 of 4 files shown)


src/controls/vr/mappings.ts
```diff
@@ -64,6 +67,28 @@ export function applyMappings(
...
     const { controlsState } = game;
     const { components } = motionController;
+
+    const listBehaviours = [
+        BehaviourMode.NORMAL,
+        BehaviourMode.ATHLETIC,
+        BehaviourMode.AGGRESSIVE,
+        BehaviourMode.DISCRETE,
+    ];
+    const hasProtoPack = game.getState().flags.quest[LBA2Items.PROTO_PACK] ||
```
agrande, line 77: That's the same index for LBA1&2, right? Otherwise this would evaluate to true on some random item.

agrande (approved):
Just tried it, looks very promising!

xesf:
True, forgot about that. There is another place with the same issue.
I will push the fixes tomorrow.

xesf:
> Just tried it, looks very promising!

Yeah, not bad. I did like the prison scene.

xesf:
I've pushed the changes.

We need to refactor this game specific stuff that is starting to be spread all over the game.

PR #574, opened 2021-05-04
Include y axis during distance checks by using threejs distance function

xesf:
**Preview here:** https://pr-574.lba2remake.net

### DIFF
--- a/src/game/Actor.ts
+++ b/src/game/Actor.ts
@@ -4,7 +4,7 @@ import {cloneDeep} from 'lodash';
 import { loadModel, Model } from '../model';
 import { loadAnimState, resetAnimState, updateKeyframeInterpolation, updateKeyframe } from '../model/animState';
 import { AnimType } from './data/animType';
-import { angleToRad, distance2D, angleTo, getDistanceLba, WORLD_SCALE } from '../utils/lba';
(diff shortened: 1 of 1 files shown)


agrande:
What prompted you to make this change?

xesf:
> What prompted you to make this change?

On scenes where you have floors, the distance was getting calculated in 2D only for different floors.

agrande:
Do you have examples so I can test?
This is a pretty big change as this command is used everywhere, and I seem to remember that we changed this back-and-forth in the past (but maybe I recall wrongly), so I want to test it thoroughly to make sure it doesn't break other scenes.

xesf:
> Do you have examples so I can test?
> This is a pretty big change as this command is used everywhere, and I seem to remember that we changed this back-and-forth in the past (but maybe I recall wrongly), so I want to test it thoroughly to make sure it doesn't break other scenes.

Start LBA1 in the Prison.
Pass the first gate towards getting the holomap and id card. Stay there for awhile until the nurse in the upper floor walks.
The nurse in the upper floor will act has you are near by when you are standing below him.

I've tested a bit and seem to have not broken other commands and distance is working as expected. Other commands in the script seems to be working fine.
Perhaps we need to test more to be sure.

agrande:
I just tested quite a bit on my own, and it seems fine.

agrande (approved):

PR #582, opened 2021-05-07
Fix for crash when loading a scene twice

agrande:
The crash was due to the cached version of the model being broken due to moving its content to another scene graph (a node can only have one parent).
I'm only caching at the HTTP level to fix this. Not optimal, but this will do for now.

**Preview here:** https://pr-582.lba2remake.net


xesf (approved):
Great to see only code been removed :)

PR #591, opened 2021-05-09
Zone Fixes

xesf:
* Fixed actors zone detection that were using BB half height position by mistake. Not all zones are guarantee to have the height of half of the BB of actors. This was more evident on LBA1 where zones are smaller than LBA2.
* Ladders on LBA2 have info prop set to one, so to maintain the same behaviour on LBA1, the info prop is been overridden during scene parsing.
* POS_POSITION command was wrongly setting the object position and was not working for sprite actors.

**Preview here:** https://pr-591.lba2remake.net

### DIFF
--- a/src/game/loop/zones.ts
+++ b/src/game/loop/zones.ts
@@ -60,15 +60,14 @@ export function processZones(game: Game, scene: Scene, time: Time) {
     const hero = scene.actors[0];
     const pos = hero.physics.position.clone();
     game.controlsState.camZone = null;
-    pos.y += 0.005;
     for (const zone of scene.zones) {
         if (zone.props.type === 2)
             continue;
 
         const box = zone.props.box;
-        if (pos.x > box.xMin && pos.x < box.xMax &&
-            pos.y > box.yMin && pos.y < box.yMax &&
-            pos.z > box.zMin && pos.z < box.zMax) {
+        if (pos.x >= box.xMin && pos.x < box.xMax &&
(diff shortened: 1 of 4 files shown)


slwilliams (approved):

src/game/loop/zones.ts
```diff
@@ -60,15 +60,14 @@ export function processZones(game: Game, scene: Scene, time: Time) {
...
     for (const zone of scene.zones) {
         if (zone.props.type === 2)
             continue;
 
         const box = zone.props.box;
-        if (pos.x > box.xMin && pos.x < box.xMax &&
-            pos.y > box.yMin && pos.y < box.yMax &&
-            pos.z > box.zMin && pos.z < box.zMax) {
+        if (pos.x >= box.xMin && pos.x < box.xMax &&
+            pos.y >= box.yMin && pos.y <= box.yMax &&
```
agrande, line 69: Why <= on y axis and not the other axis?

agrande (commented):
Can't test this right now, but this is an impactful change, there are a few scenes in LBA2 where I think it could break things. I'll check it tonight.

src/game/loop/zones.ts
xesf: To keep it consistent with the original.

Please do, let me know if you find anything. I've been testing quite a lot different scenes on both games, but there seem to always been edge cases.

agrande:
Ok, I tested it a bit and all seems to work as expected. 👍

PR #592, opened 2021-05-09
Fixed go to positions commands to wait for an actor to finish talking

xesf:
An example to test this are the platforms which were still going to their position even during actor interactions.

**Preview here:** https://pr-592.lba2remake.net

### DIFF
--- a/src/game/scripting/move.ts
+++ b/src/game/scripting/move.ts
@@ -8,6 +8,11 @@ export function GOTO_POINT(this: ScriptContext, point: Point) {
     if (!point) {
         return;
     }
+    if (this.game.getState().actorTalking > -1) {
+        this.state.reentryOffset = this.state.offset;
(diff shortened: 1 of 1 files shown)


agrande:
Isn't there a risk that actors will keep moving forward without turning, and missing their target, if their animation was set to the forward animation before running this command?

xesf:
> Isn't there a risk that actors will keep moving forward without turning, and missing their target, if their animation was set to the forward animation before running this command?

This could potentially happen for the actor that is talking, but unlikely as interaction will most likely reset to a different track.
We enable animations to be playing for who is talking, so we can see actors turning and have Twinsen with reading/talking animation.

All other actors that are not talking will be blocked and only the scripts will be executed.

agrande:
Makes sense.

agrande (approved):

PR #602, opened 2021-05-17
Radar SFX

agrande:
**Preview here:** https://pr-602.lba2remake.net


src/game/scenery/isometric/metadata/fx/Fx.ts
```diff
@@ -0,0 +1,19 @@
...
+    paletteTexture: THREE.DataTexture;
+    lutTexture: THREE.DataTexture;
+    light: THREE.Vector3;
+}
+
+export interface FxConstructor {
+    new(props: FxProps): Fx;
+}
+
+export interface Fx {
```
xesf, line 16: Looks more like a GameComponent interface and not so Fx related.

xesf (approved):

src/game/scenery/isometric/metadata/fx/Fx.ts
agrande: Yeah, now that you mention it, it looks like it could be generalized and used in other places. I think I'll wait for the need to arise for doing it though.

PR #612, opened 2021-05-21
Fix anim interpolation

agrande:
This fixes the animation interpolation.
The key was to use the current skeleton state at the moment we start interpolating instead of the keyframe from the previous anim.
I refactored the code pretty deeply in the process to use classes with as many type annotations as possible so we can navigate around the code more easily and be more confident of not breaking things when changing it.
Seems like it improves the tunic protective glass animation in the museum to some extent, although there’s still something not working properly there.
Also, I’m using quaternions in the bones instead of eulers for smoother/cheaper interpolation, I moved the conversion to quaternions to the parser.

**Preview here:** https://pr-612.lba2remake.net


xesf:
Tested and reviewed and all looks fine for both games.

I notice that interpolation is not happening when you switch behaviours, is that intentional?

agrande:
> I notice that interpolation is not happening when you switch behaviours, is that intentional?

Are you sure it's not? It looked like it was to me.

agrande:
> > I notice that interpolation is not happening when you switch behaviours, is that intentional?
> 
> Are you sure it's not? It looked like it was to me.

Did you switch through the behavior menu or the shortcuts?

slwilliams:
yea this is so much better. One bug I did find is that if you drown we get an undefined error likely due to setFromJSON or similar.

agrande:
> yea this is so much better. One bug I did find is that if you drown we get an undefined error likely due to setFromJSON or similar.

Ok, I'll check that.

agrande:
> I notice that interpolation is not happening when you switch behaviours, is that intentional?

Ok, I looked at this again, and you're right. It happens because we're using another entity for different behaviours, and so the anim is completely reset. I'll see if I can do something about it.

agrande:
> yea this is so much better. One bug I did find is that if you drown we get an undefined error likely due to setFromJSON or similar.

Fixed.

slwilliams (approved):

xesf (approved):

PR #615, opened 2021-05-23
Revert logic for anim ended callback

slwilliams:
It seems hasEnded doesn't actually trigger when the animation has completely played through once which is what I want to trigger the callback being fired. This reverts to the logic we have in the previous implementation and shouldn't break anything since _hasEnded remains untouched.

At the moment falling for example is broken in autopush, this fixes that.

**Preview here:** https://pr-615.lba2remake.net

### DIFF
--- a/src/model/anim/AnimState.ts
+++ b/src/model/anim/AnimState.ts
@@ -75,7 +75,7 @@ export default class AnimState {
             this.pose.setFromKeyFrames(this.kfs, this.currentTime);
         }
         this.pose.skeleton.updateHierarchy();
-        if (this._hasEnded && this.onAnimEnd) {
+        if (this._currentFrame === this.anim.numKeyframes - 1 && this.onAnimEnd) {
             this.onAnimEnd();
             this.onAnimEnd = null;
         }
--- a/src/game/Actor.ts
+++ b/src/game/Actor.ts
@@ -613,6 +613,10 @@ export default class Actor {
             return;
         }
         this.props.animIndex = index;
+        if (this.animState.onAnimEnd) {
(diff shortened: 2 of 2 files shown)


xesf (approved):

src/model/anim/AnimState.ts
```diff
@@ -75,7 +75,7 @@ export default class AnimState {
             this.pose.setFromKeyFrames(this.kfs, this.currentTime);
         }
         this.pose.skeleton.updateHierarchy();
-        if (this._hasEnded && this.onAnimEnd) {
+        if (this._currentFrame === this.anim.numKeyframes - 1 && this.onAnimEnd) {
```
agrande, line 78: I'm not sure I completely understand what's going on here.
What exactly is the difference and how does it affect falling?

slwilliams: so the callbacks are supposed to be executed after a single play through of the animation. hasEnded is set true if the nextkeyframe is the looping keyframe. This isn't the same thing as the animation having played through the entire way e.g. the loopframe could be the second frame (in the case of falling) in which case we execute the callback way before the animation has completed playing. By triggering the callback on the end frame we get what we want.

agrande: Ok, so I'm reading again the condition that sets `_hasEnded` to `true`, and it's indeed very wrong. I think I should fix that instead of having distinct logic for hasEnded and triggering the callback. Those are supposed to happen at the same time.
Also, with your code, I think the callback is in fact called before the end of the anim. The anim ends when we finish interpolating between the last keyframe and the loop frame (or first frame if loop frame is not set).

PR #617, opened 2021-05-23
Lba1 fixes

xesf:
Lba1 fixes:
* Remove default lba2 game flags
* While other actors were talking, twinsen animation was reseted, causing scripts to detect the animation change.
* Reducing the LBA1 models bounding boxes, which unblocks loads of collision issues with BB that are not squared.
* Adjusted VR camera height for Twinsen on LBA1

**Preview here:** https://pr-617.lba2remake.net


agrande (approved):
Looks ok to me.

PR #625, opened 2021-05-28
Fix extras not being created when actors die

slwilliams:
**Preview here:** https://pr-625.lba2remake.net


agrande (approved):
Do I need to do something to make this work with the magicball too?

slwilliams:
nope everything will just work since this is encapsulated in the hit logic.

PR #628, opened 2021-05-29
VR: Hero walking to point

agrande:
Twinsen now walks automatically to points and the view adjusts itself every 2 seconds.
Also allowing to rotate the view during cinematic dialogues.

This fixes exiting the Tralü's cave, as well as entering Twinsen's house in LBA1.

**Preview here:** https://pr-628.lba2remake.net


xesf (approved):
Haven't tested yet. I will try tomorrow morning.

PR #635, opened 2021-06-01
Push Objects

xesf:
Allow Twinsen to push objects like the Warehouse Crates in LBA1.

**Preview here:** https://pr-635.lba2remake.net


slwilliams:
Hmm this seems to allow pushing things that shouldn't be able to be pushed. Try the protection spell cave with those boxes you can push, you can also push the blue rocks which you shouldn't be able to.

xesf:
Thanks for looking.
Weird bug, I will need to check the actors props and see what's going on.

agrande:
Nice work.
An improvement that we'll need for LBA2 is the ability to push the boxes sideways.
Right now it's only axis-aligned movement it seems.

xesf:
Totally forgot it was a thing in LBA2 :D

slwilliams:
ah ok just tried to complete the warehouse puzzle in lba 1 and you can push the crates inside of each other i.e. we need collision detection between pushable objects.

xesf:
That means I broke it again with my changes from yesterday to fix the sprites collision in the cave, because the crates were colliding with each other.
Tricky to get this working properly :)

xesf:
Looks like it works fine for both games now.
LBA2 still missing the side push, but lets do that one later.

PR #644, opened 2021-09-28
Improve Citadel Island metadata

fraserjgordon:
Most of the game vars that are referenced from any of the Citadel Island scenes now have names and enumeration values defined. There are also some related scene vars that have been filled out too. Finally, the Pearl Cave "CloverBox" actor has been renamed to "GiantClam" as the scripting indicates that that is its role.

**Preview here:** https://pr-644.lba2remake.net

### DIFF
--- a/www/metadata/lba2/game.json
+++ b/www/metadata/lba2/game.json
@@ -209,7 +209,10 @@
         "2": "removed_from_holomap"
       }
     },
-    null,
+    {
+      "name": "baggage_claim_downstairs_door_opened",
+      "type": "boolean"
+    },
     null,
     {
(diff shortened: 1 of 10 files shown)


xesf:
Hi @fraserjgordon 

Thanks for the PR, we appreciate the hard work on setting up the metadata.

I will have a look during the day and let you know.

Btw, if you are interested to do the same for LBA1 that would be brilliant, most of the metadata is still missing.

Cheers

agrande (approved):
Thanks for the PR, great work!

xesf (approved):

PR #645, opened 2021-09-28
Implement the 'set_sprite' opcode

fraserjgordon:
This updates the sprite of the actor to the one with the given index.

To test, go to Twinsen's neighbour's house and set the value of game var 66 (with #644  it is called 'sewer_map_state') to a value >= 1 (`GALLIC_ACID_USED`) and the map on the desk will immediately update with the new sprite. To go back to the original sprite, the var needs to be set to 0 and the scene exited (as the life script for the map stops running after updating the sprite).

This also works for the various switches in the world (e.g. in Lupin-Bourg's sewers).

**Preview here:** https://pr-645.lba2remake.net

### DIFF
--- a/src/game/scripting/life.ts
+++ b/src/game/scripting/life.ts
@@ -609,7 +609,9 @@ export function ACTION(this: ScriptContext) {
 
 export const SET_FRAME = unimplemented();
 
-export const SET_SPRITE = unimplemented();
+export function SET_SPRITE(this: ScriptContext, index: number) {
+    this.actor.setSprite(this.scene, index);
+}
 
 export function SET_FRAME_3DS(this: ScriptContext) {
 


src/game/scripting/life.ts
```diff
@@ -609,7 +609,10 @@ export function ACTION(this: ScriptContext) {
 
 export const SET_FRAME = unimplemented();
 
-export const SET_SPRITE = unimplemented();
+export function SET_SPRITE(this: ScriptContext, id) {
+    this.actor.props.spriteIndex = id;
+    this.actor.reloadModel(this.scene);
```
fraserjgordon, line 614: I know nothing about async JS programming so I'm not sure if calling the async `reloadModel` method like this is sound but it does seem to work in my testing...

xesf: That is fine, we don't need to wait for the reload to proceed with the script.

You should await if you want to wait for the results of the async function, but also it is require to have at least 1 await inside an async function (in this case set_sprite is not async.

xesf, line 614: In terms of the code, you can actually use [setSprite](https://github.com/agrande/lba2remake/blob/308654bcd038447a1900c45d7f06fe41962be965/src/game/Actor.ts#L623) from the actor itself so you won't duplicate the code. There are extra validations there as well so it will be safer to use it instead.

xesf (changes requested):

src/game/scripting/life.ts
```diff
@@ -609,7 +609,10 @@ export function ACTION(this: ScriptContext) {
 
 export const SET_FRAME = unimplemented();
 
-export const SET_SPRITE = unimplemented();
+export function SET_SPRITE(this: ScriptContext, id) {
```
agrande, line 612: Maybe add a type annotation on the id param.
Not strictly needed, but preferable in general to add type to function params.

agrande (approved):
Lgtm, besides the change @xesf suggested.

src/game/scripting/life.ts
fraserjgordon: Done

src/game/scripting/life.ts
fraserjgordon: Done.

xesf (approved):

PR #646, opened 2021-09-29
Implement var_game_to_track and track_to_var_game

fraserjgordon:
These opcodes load/store the current move track of an actor from/to a game var. Also check these opcodes when building the "only in scene" game var list in the editor.

The main use of these opcodes seems to be for implementing timers. For example, most (all?) areas on Desert Island contain an invisible actor (e.g. actor_19 in the School of Magic Classroom) that has a move script which has a chain of tracks and waits and a life script that continually saves the current track to a game var (var_80). This var is then used to drive the state of the Peddler as he travels around the island.

**Preview here:** https://pr-646.lba2remake.net

### DIFF
--- a/src/game/scene/variables.ts
+++ b/src/game/scene/variables.ts
@@ -26,7 +26,9 @@ export function findUsedVarGames(actors: Actor[]) {
         const commands = actor.scripts.life.commands;
         for (const cmd of commands) {
             let value = null;
-            if (cmd.op.command === 'SET_VAR_GAME') {
+            if (cmd.op.command === 'SET_VAR_GAME'
+                || cmd.op.command === 'TRACK_TO_VAR_GAME'
+                || cmd.op.command === 'VAR_GAME_TO_TRACK') {
                 value = cmd.args[0].value;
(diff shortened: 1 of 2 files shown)


xesf (approved):

src/game/scene/variables.ts
```diff
@@ -26,7 +26,9 @@ export function findUsedVarGames(actors: Actor[]) {
         const commands = actor.scripts.life.commands;
         for (const cmd of commands) {
             let value = null;
-            if (cmd.op.command === 'SET_VAR_GAME') {
+            if (cmd.op.command === 'SET_VAR_GAME'
+                || cmd.op.command === 'TRACK_TO_VAR_GAME'
+                || cmd.op.command === 'VAR_GAME_TO_TRACK') {
```
agrande, line 31: Nice that you added those here!
We're probably also missing some other commands that use VAR_GAMEs.

agrande (approved):
Nice to see the peddler working now.
I'm not sure how we'll manage to make him go from one scene to another seemlessly though (and also appear in only one scene at once).
Any idea what actor_19 is for?

PR #647, opened 2021-09-29
Implement the 'no_shock' opcode

fraserjgordon:
This opcode blocks any hit processing when enabled for an actor (no damage animations are played; no damage is taken; and hit_by is not set).

### DIFF
--- a/src/game/Actor.ts
+++ b/src/game/Actor.ts
@@ -119,6 +119,7 @@ export interface ActorState {
     floorSound: number;
     floorSound2: number;
     nextAnim: number;
+    noShock: boolean;
 }
 
 export enum SlideWay {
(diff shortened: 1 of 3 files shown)


src/game/Actor.ts
```diff
@@ -822,6 +828,7 @@ export default class Actor {
             floorSound: -1,
             floorSound2: -1,
             nextAnim: null,
+            noShock: false,
```
xesf, line 831: Looks like we have a noElectricShock static flag that is set in the Scene for each actor. Perhaps these two should be merged?

This is something we have to improve overall in the project, we started using flags and got mixed up with state.

xesf (changes requested):

src/game/Actor.ts
fraserjgordon: I think (but am not 100% sure - I'll need to do some more digging) that the flags are different. `noElectricShock` might be related to whether the model has an "electric shock x-ray" model.

fraserjgordon: I checked this and it does seem like both flags are the same (and noElectricShock is just misnamed). I'll fold them together. Might be worth keeping both tracked though (the value at load time and the value as modified by scripts).

PR #648, opened 2021-09-29
Implement armour and related opcodes

fraserjgordon:
Armour provides a flat reduction to incoming damage. If all the damage is blocked, the hit animations and logic still play but no life points are subtracted.

The LBA2 data files tend to set an armour value of 51 for unkillable actors. No special coding for this value has been added; magic ball damage tops out at 40 so such actors are functionally invincible.

**Preview here:** https://pr-648.lba2remake.net

### DIFF
--- a/src/game/Actor.ts
+++ b/src/game/Actor.ts
@@ -43,6 +43,7 @@ export interface ActorProps {
     sceneIndex: number;
     pos: [number, number, number];
     life: number;
+    armour: number;
     flags: ActorFlags;
     entityIndex: number;
     bodyIndex: number;
@@ -678,12 +679,13 @@ export default class Actor {
         }
 
         let life = -1;
-        // TODO(scottwilliams): This doesn't take into account actor armour.
+        const armour = this.props.armour;
+        const damage = Math.max(0, hitStrength - armour);
         if (this.index === 0) {
-            this.game.getState().hero.life -= hitStrength;
+            this.game.getState().hero.life -= damage;
             life = this.game.getState().hero.life;
         } else {
-            this.props.life -= hitStrength;
+            this.props.life -= damage;
             life = this.props.life;
         }
 
@@ -887,6 +889,7 @@ export function createNewActorProps(
         index: scene.actors.length,
         pos: details.position.toArray() as any,
         life: 255,
+        armour: 0,
         flags: {
             hasCollisions: true,
             hasCollisionBricks: true,
--- a/src/game/scripting/life.ts
+++ b/src/game/scripting/life.ts
@@ -447,9 +447,8 @@ export function SUB_LIFE_POINT_OBJ(this: ScriptContext, actor: Actor, value) {
     }
 }
 
-export function HIT(this: ScriptContext, actor: Actor, strength) {
-    actor.state.wasHitBy = this.actor.index;
-    actor.props.life -= strength;
+export function HIT(this: ScriptContext, actor: Actor, strength: number) {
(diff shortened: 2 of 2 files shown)


fraserjgordon:
I've done a bit of research on weapon damage vs armour in original LBA2 this evening. The zero-armour damage for everything is:

Yellow magic: 10
Green magic: 20
Red magic: 30
Fire magic: 40
Darts: 8
Sword: 35
Blowtron: 23
Blowgun: 15
Laser pistol: 30 (plus 12 for each the impact fragments)

The jumping about sword attack (in athletic mode) does 3x the amount of damage, but this is calculated after armour is subtracted. So it does 3x35=105 damage at 0 armour, 3x(35-34)=3 damage at 34 armour and 3x(35-35)=0 damage at 35+ armour.

To get these numbers, I used LBArchitect to add the following to the main behaviour of Mr Bazoo:
```
  GIVE_GOLD_PIECES 500 
  STATE_INVENTORY 23 1
  STATE_INVENTORY 9 2
  SET_VAR_GAME 2 3
  SET_VAR_GAME 9 1 
  SET_VAR_GAME 10 1 
  SET_VAR_GAME 11 1 
  SET_VAR_GAME 23 1 
  SET_VAR_GAME 46 1 
  SET_VAR_GAME 50 1
  SET_MAGIC_LEVEL 4
  SET_LIFE_POINT_OBJ SELF 100
  SET_ARMOUR 0
```
Then the following in behaviour 1:
```
IF LIFE_POINT < 100 
    ADD_LIFE_POINT_OBJ SELF 1 
    ADD_GOLD_PIECES 1 
ENDIF 
```
Then, after you attack the poor Grobo, you can check Twinsen's kashes to see how much damage each attack did.

xesf (approved):
Thanks for the detailed info, very useful.
I wonder if we should save those type of things somewhere in the repo, like a knowledge based. @fraserjgordon @agrande @slwilliams any thoughts?

slwilliams:
Changes looks good, thanks for implementing this!

Re knowledge base I still use http://lbafileinfo.kaziq.net/ might be good to add a new page there?

agrande:
> To get these numbers, I used LBArchitect to add the following to the main behaviour of Mr Bazoo:
> 
> ```
>   GIVE_GOLD_PIECES 500 
>   STATE_INVENTORY 23 1
>   STATE_INVENTORY 9 2
>   SET_VAR_GAME 2 3
>   SET_VAR_GAME 9 1 
>   SET_VAR_GAME 10 1 
>   SET_VAR_GAME 11 1 
>   SET_VAR_GAME 23 1 
>   SET_VAR_GAME 46 1 
>   SET_VAR_GAME 50 1
>   SET_MAGIC_LEVEL 4
>   SET_LIFE_POINT_OBJ SELF 100
>   SET_ARMOUR 0
> ```
> 
> Then the following in behaviour 1:
> 
> ```
> IF LIFE_POINT < 100 
>     ADD_LIFE_POINT_OBJ SELF 1 
>     ADD_GOLD_PIECES 1 
> ENDIF 
> ```
> 
> Then, after you attack the poor Grobo, you can check Twinsen's kashes to see how much damage each attack did.

Interesting to see how you did this!

About the knowledge base, I agree with @slwilliams that it would be nice to centralize the info where the community already goes (http://lbafileinfo.kaziq.net). Would be nice to update / create new pages with the information we have. I updated some pages a few years ago with some stuff we discovered while working on the remake, but there's a lot more we know about that is not there.

PR #649, opened 2021-10-01
Inventory state tracking

fraserjgordon:
In addition to using global vars to track items in Twinsen's inventory, each inventory slot also has a "state". This is used to determine which variant of an item Twinsen holds. The affected items are:

- robes vs wizard robes
- blowgun vs blowtron
- protopack vs jetpack
- etc

These changes implement support for tracking that state. Individual commit messages have details.

**Preview here:** https://pr-649.lba2remake.net

### DIFF
--- a/src/game/GameState.ts
+++ b/src/game/GameState.ts
@@ -5,6 +5,7 @@ import { getParams } from '../params';
 import Actor from './Actor';
 import AnimState from '../model/anim/AnimState';
 import { AnimStateJSON } from '../model/anim/types';
+import { GetInventorySize, LBA2Items } from './data/inventory';
 
 export interface GameConfig {
     displayText: boolean;
@@ -92,6 +93,7 @@ export function createGameState(): GameState {
         actorTalking: -1,
         flags: {
             quest: createQuestFlags(),
+            inventory: createInventoryFlags(),
             holomap: createHolomapFlags()
         },
         save(hero: Actor) {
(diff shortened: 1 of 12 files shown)


src/game/GameState.ts
```diff
@@ -184,4 +195,10 @@ export function setMagicBallLevel(state: GameState, index: number) {
 
     state.hero.magicball = magicball;
     state.hero.handStrength = handStrength;
+
+    // Update the magic ball and ring of lightning item states.
+    state.flags.inventory[LBA2Items.MAGIC_BALL] = index;
```
xesf, line 200: Since this line is not protected by the game condition I would say to use
GetInventoryItems().MAGIC_BALL instead of LBA2Items.MAGIC_BALL

src/game/data/inventory.ts
```diff
@@ -7,11 +7,13 @@ import { getParams } from '../../params';
 const isLBA1 = getParams().game === 'lba1';
 
 const LBA2InventoryColumns = 7;
-const LBA2InventoryRows = 5;
+const LBA2InventoryRows = 6;    // Original: 5, but it overloaded some slots.
```
xesf, line 10: I like the idea of been extended. We no longer have the screen space constrain no to expand it.

xesf (approved):

xesf:
btw @fraserjgordon you can merge the code yourself as long you have at least 1 approval.

PR #651, opened 2021-10-02
Slightly improve sprite rendering

fraserjgordon:
This commit adds some additional cases for how we turn sprites into 3D objects: as well as the existing XY and ZY plane meshes, this adds XZ plane meshes and a fallback isometric-aligned mesh.

Some objects are still drawn poorly (particularly in the fallback case) but they tend to be better than before.

**Preview here:** https://pr-651.lba2remake.net

### DIFF
--- a/src/game/scenery/isometric/sprites.ts
+++ b/src/game/scenery/isometric/sprites.ts
@@ -99,22 +99,81 @@ export async function loadSprite(
 }
 
 function loadMesh(index, sprite, box) {
+    // flat - draw the sprite on the XZ plane.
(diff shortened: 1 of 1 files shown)


xesf (approved):
This is an awesome fix!! We have this broken for so long...

Few things I notice though. The sprites don't seem to be positioning correctly, their pivot point does not seem to be the centre of the sprite.
For example in LBA1 prison scene, you can see both the platform and the level above just slightly off to the left/down side. Very tiny though, looks like its a half a brick.

Anyway this is great to finally have these sprites look at they should.

PR #652, opened 2021-10-04
Implement loading (but not animation) of Anim3DS

fraserjgordon:
The ANIM3DS.HQR file stores these sprites as a series of frames and an info block giving the frame range of each animation. The sprites themselves are normal sprites.

This change ensures that actors with the Anim3DS flag load the right sprites but they are not actually animated at this point. For now, we just render the first sprite in the sequence. To see an example, visit the Sendell's Ball room in the sewers.


slwilliams (approved):
Looks good, although I was expecting this to also fix the sprites in other places e.g. temple of Bu, guess that's another issue: 
image

xesf:
It looks good indeed.

While testing I notice another bug. (not related)
So now that you have implemented the set sprite and correctly display the sprite, we can now see the Sendell sign in the sewers just before entering the Sendell's Ball room. The animation is done via script, but for some reason if we go into In-Game menu (ESC key) and go back to it, the move script no longer runs. This may be a major bug causing loads of other sort of scripting issues if we pause the game.

xesf:
So LBA1 crashes after this change because it does not know the new Resources for ANIM3DS and ANIM3DS_INFO.
The reference should be in the LBA1 resource map as well, even if does not have a path (see BODY_TEXTURE).

xesf (changes requested):
So LBA1 crashes after this change because it does not know the new Resources for ANIM3DS and ANIM3DS_INFO.
The reference should be in the LBA1 resource map as well, even if does not have a path (see BODY_TEXTURE).

fraserjgordon:
@slwilliams I think the problem might be that the various script opcodes for the animated sprites aren't implemented yet (they're on my todo list). I'll take a look though in case it is something else.

@xesf thanks for that info! I've been concentrating my testing on LBA2 for now as I've never really played LBA1 (though this might finally motivate me to do it!)

PR #654, opened 2021-10-06
Improve handling of zone flags

fraserjgordon:
This PR starts with a commit that reworks how we handle zone flags (by splitting them out from the `infoX` members into more friendly named flags). It then implements a number of life script opcodes used for enabling and disabling zones.

**Preview here:** https://pr-654.lba2remake.net


fraserjgordon:
I've not implemented the `SET_GRM` opcode in this PR - it is used to enable or disable scene fragments. As it has an immediate rendering effect, I thought it would be better to leave it unimplemented until it gets plumbed into the renderer rather than just uselessly toggling the bit.

Also, an question of style preferences - should the other zone opcodes (`ESCALATOR` and `LADDER`) be renamed to follow the pattern of the rest of them (so `SET_ESCALATOR` and `SET_LADDER`)? Or even do a wider rename:
- `SET_CAMERA_ZONE`
- `SET_TELEPORT_ZONE`
- `SET_ESCALATOR_ZONE`
- `SET_LADDER_ZONE`
- `SET_TRAP_ZONE`

src/game/Zone.ts
```diff
@@ -146,4 +216,88 @@ export default class Zone {
         this.labelCtx.fillText(this.name, 128 + 18, 38, 256 - 64);
         this.labelTexture.needsUpdate = true;
     }
+
+    private unpackProps() {
```
xesf, line 220: I wonder we can spread this to main zone object depending on the type?! instead of having a separate prop for each?
Perhaps a zone type in typescript for each so it can be cast while referencing in other parts of the engine.

I like idea of having more specific meaning for fields in each type, but at the same time does not make sense to have another depth inside zones when it should be the same thing.

Anyway, I don't think this blocks de PR from been merged

xesf (approved):
> I've not implemented the `SET_GRM` opcode in this PR - it is used to enable or disable scene fragments. As it has an immediate rendering effect, I thought it would be better to leave it unimplemented until it gets plumbed into the renderer rather than just uselessly toggling the bit.
> 
> Also, an question of style preferences - should the other zone opcodes (`ESCALATOR` and `LADDER`) be renamed to follow the pattern of the rest of them (so `SET_ESCALATOR` and `SET_LADDER`)? Or even do a wider rename:
> 
> * `SET_CAMERA_ZONE`
> * `SET_TELEPORT_ZONE`
> * `SET_ESCALATOR_ZONE`
> * `SET_LADDER_ZONE`
> * `SET_TRAP_ZONE`

I am fine with those changes.

agrande (approved):
Thanks for the edits!
Great work! 😊

